PR #14, opened 2023-10-03
Preliminary new routines for fitting and AoE calibration

theHenks:
Preliminary fitting and routines are working and stable, but need review. Especially, the combined fitting and AoE routines. Also, need more routines for the various drift time corrections and their routines. Some scripts are available but they need some more testing.
@oschulz Could you also check the `Project.toml` again here?


Project.toml
```diff
@@ -5,30 +5,41 @@ version = "0.1.0"
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
 ArraysOfArrays = "65a8f2f4-9b39-5baf-92e2-a9cc46fdf018"
+BAT = "c0cd4b16-88b7-57fa-983b-ab80aecada7e"
```
oschulz, line 8: If there's a package extension for BAT, then BAT should be in `[weakdeps]`.

Project.toml
```diff
@@ -5,30 +5,41 @@ version = "0.1.0"
...
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+Formatting = "59287772-0a20-5a39-b81b-1366585eb4c0"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 IrrationalConstants = "92d709cd-6900-40b7-9082-c6be49f344b6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LinearRegression = "92481ed7-9fb7-40fd-80f2-46fd0f076581"
+LsqFit = "2fda8390-95c7-5789-9bda-21331edee243"
+Optim = "429524aa-4258-5aef-a3af-852621145aeb"
```
oschulz, line 21: Optim should either be a hard dependency with Pkg ext, or a weak dependency with Pkg ext.

oschulz:
Look good to me in general (except for the test failures ;-) ).

oschulz:
Let's just require Julia v1.9 (in deps and CI).

Project.toml
theHenks: I think we don't need the extension right now. So I would just delete the extension and just leave BAT as `[deps]` as long as we can not use the DistributionTransform separately without `BAT`.  And if we wanna go to Bayesian fits at some point we anyhow need it.

Project.toml
theHenks: I would then make it a hard one, since we need it all the time.

Project.toml
oschulz: Ok

Project.toml
oschulz: Ok

PR #28, opened 2023-12-11
Improve binning and fitting of A/E histograms

fhagemann:
Outliers at high A/E values would shift the A/E maximum-likelihood fits towards higher A/E values.
Therefore, the binning of the A/E should be chosen such that these outliers are excluded when fitting Compton bands.

Here, instead of selecting the bin ranges based on quantiles, a first fit to the A/E peak is performed using `curve_fit` and the upper A/E bin range is set to 4σ above the main peak.

For the fitting, the likelihood is set to zero for A/E values outside of the histogram ranges.


oschulz:
@theHenks ?

theHenks:
Let's merge that PR and then all the rest of the Vancouver CM presentation for a new release. @oschulz is that ok?

oschulz:
Sure!

oschulz:
You need to resolve the merge conflicts first, though. :-)

fhagemann:
These changes are already on the `dev` branch of legend-exp/LegendSpecFits.jl, so merging `dev` into `main` should be the way to go

PR #38, opened 2024-03-12
Routine Updates

theHenks:
Updated routines for 
- A/E and PSD 
- Energy calibration
- Simple Fits 

to be able to handle `Measurements` objects and deal properly with units. 
Contains also bug fixes to various routines to improve performance.


ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -329,10 +328,8 @@ end
 
 @recipe function f(report_window_cut::NamedTuple{(:h, :f_fit, :x_fit, :low_cut, :high_cut, :low_cut_fit, :high_cut_fit, :center, :σ)})
     xlims := (value(ustrip(report_window_cut.center - 5*report_window_cut.σ)), value(ustrip(report_window_cut.center + 5*report_window_cut.σ)))
+    xlims := (ustrip(report_window_cut.center - 5*report_window_cut.σ), ustrip(report_window_cut.center + 5*report_window_cut.σ))
```
fhagemann, line 331: Can you set `xlims` twice in a recipe without causing an error?

Project.toml
```diff
@@ -9,6 +9,7 @@ BAT = "c0cd4b16-88b7-57fa-983b-ab80aecada7e"
 ChangesOfVariables = "9e997f8a-9a97-42d5-a9f1-ce6bfc15e2c0"
 DensityInterface = "b429d917-457f-4dbc-8f4c-0cc954292b1d"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
oschulz, line 12: Why do we depend on Documenter?

PR #42, opened 2024-03-18
fit calibration curve with chi2fit

LisaSchlueter:
add new file fit_calibration.jl

- fit calibration function using `chi2fit`
- to be used in `legend-julia-dataflow/process_energy_calibration.jl` like this: `fit_calibration(1,µ, pp_fit ;e_type = e_type_str)`
- returns fit `par` with uncertainties and units
- `func` (fit function with parameter values w/o uncertainties)
- `func_generic` (generic fit function w/o parameter values)


fhagemann:
Is there any way we can release a new version of LegendDataManagement with `PolCalFunc` exported such that the tests don't fail?

LisaSchlueter:
fit calibration curve with chi2fit

PR #46, opened 2024-04-03
bug fix chi2fit

LisaSchlueter:
- normalized residuals did not take x-uncertainty into account
- now fixed: combined y and x uncertainties considered.


theHenks (approved):
Fine for me

PR #52, opened 2024-04-23
Prelimenary Lq implementation

DaGeibl:
WIP LQ cut implementation
Probably could move a bit more code from my script into the package
Works like the following:
1. Load Raw LQ
2. correct either by dividing throuh e_cusp or use Compton band fits, simliar to A/E
3. Do a drift time correction
4. Do a side band subtraction on DEP
5. Fit DEP and define cut as 3 sigma of fit


src/peakshapes.jl
```diff
@@ -6,7 +6,7 @@
 Equivalent to `pdf(Normal(μ, σ), x)`
 """
 gauss_pdf(x::Real, μ::Real, σ::Real) = inv(σ * sqrt2π) * exp(-((x - μ) / σ)^2  / 2)
-
+export gauss_pdf
```
theHenks, line 9: Remove that. Otherwise it is always exported and just leads to more confusion

theHenks:
Nice, a couple comments to that:
- You changed the recipe for existing plots. We use these in the data production. So this has to go and either into a new recipe or we change that during plotting.
- For the `Normal` fit for the cut extraction, you do a binned fit without `truncating` it, as far as I see. I think it makes sense to move that into a single function and make it independent of the `LQ` part, since we gonna use that quite often and it is nicer to test. So to speak: Modify the `fit_single_trunc_gauss` function so that it also can fit binned data and returns the correct results and then call it in your 
- Remove the `export` in the `peakshapes` for the Gauss, otherwise that is confusing when you load the package
- It would be nice to also have wrapper functions for how to call all of that. So a function called `correct_lq` or something similar which does the corrections like `energy` correction and drift time correction etc, and can maybe get `kwargs` where you can set what you actually wanna do. This would help in setting things for each detector later individually. 
- Is the code for the corrections any different to the `A/E` code? If not it is a lot of duplication. Maybe then just call for now the `A/E` code

src/lqfit.jl
```diff
@@ -0,0 +1,334 @@
...
+f_lq_sig(x, v)     = aoe_compton_signal_peakshape(x, v.μ, v.σ, v.n)
+f_lq_bkg(x, v)     = aoe_compton_background_peakshape(x, v.μ, v.σ, v.B, v.δ)
+
+#######
+f_lq_mu(x, p) = p[1] .+ p[2].*x
+@. f_lq_sigma(x, p) = sqrt(abs(p[1]) + abs(p[2])/x^2)
+#######
+
+# lq compton centroids energy depencence
+MaybeWithEnergyUnits = Union{Real, Unitful.Energy{<:Real}}
```
theHenks, line 12: This is doubled with the `A/E` code, can go

src/lqfit.jl
```diff
@@ -0,0 +1,334 @@
...
+Estimate peak parameters for a single peak in a histogram using the maximum, the FWHM and the area of the peak.
+
+# Returns
+    * `peak_pos`: Position of the peak
+    * `peak_fwhm`: Full width at half maximum of the peak
+    * `peak_sigma`: Standard deviation of the peak
+    * `peak_counts`: Counts of the peak
+    * `mean_background`: Mean background of the peak
+"""
+function estimate_single_peak_stats_psd_lq(h::Histogram{T}) where T<:Real
```
theHenks, line 34: Is this any differen to the `A/E` code? if not, just call the code for that here.
``` julia
function estimate_single_peak_stats_psd_lq(h::Histogram{T}) where T<:Real             estimate_single_peak_stats_psd(h)
end
```

src/lqfit.jl
```diff
@@ -0,0 +1,334 @@
...
+    * `min_lq`: Array of minimum A/E values for each compton band
+    * `max_lq`: Array of maximum A/E values for each compton band
+    * `mean_peak_pos`: Mean peak position of all compton bands
+    * `std_peak_pos`: Standard deviation of the peak position of all compton bands
+    * `simple_pars_lq_μ`: Simple curve fit parameters for the peak position energy depencence
+    * `simple_pars_error_lq_μ`: Simple curve fit parameter errors for the peak position energy depencence
+    * `simple_pars_lq_σ`: Simple curve fit parameters for the peak sigma energy depencence
+    * `simple_pars_error_lq_σ`: Simple curve fit parameter errors for the peak sigma energy depencence
+"""
+function generate_lq_compton_bands(lq::Vector{<:Real}, e::Vector{<:T}, compton_bands::Vector{<:T}, compton_window::T) where T<:Unitful.Energy{<:Real}
```
theHenks, line 91: Same here as above

src/lqfit.jl
```diff
@@ -0,0 +1,334 @@
...
+"""
+    fit_single_lq_compton(h::Histogram, ps::NamedTuple{(:peak_pos, :peak_fwhm, :peak_sigma, :peak_counts, :mean_background, :μ, :σ), NTuple{7, T}}; uncertainty::Bool=true) where T<:Real
+
+Perform a fit of the peakshape to the data in `h` using the initial values in `ps` while using the `f_lq_compton` function consisting of a gaussian SSE peak and a step like background for MSE events.
+
+# Returns
+    * `result`: NamedTuple of the fit results containing values and errors
+    * `report`: NamedTuple of the fit report which can be plotted
+"""
+function fit_single_lq_compton(h::Histogram, ps::NamedTuple; uncertainty::Bool=true)
```
theHenks, line 233: Did you play with the `priors` or are these also the same ones? Maybe also call the `A/E` routines here.

src/lqcut.jl
```diff
@@ -0,0 +1,133 @@
...
+"""
+    fit_lq_DEP_ampl(h::Histogram, ps::NamedTuple{(:peak_pos, :peak_fwhm, :peak_sigma, :peak_counts, :mean_background, :μ, :σ), NTuple{7, T}}; uncertainty::Bool=true) where T<:Real
+
+Perform a fit of the peakshape to the data in `h` using the initial values in `ps` while using the `f_lq_DEP` function consisting of a gaussian peak.
+
+# Returns
+    * `result`: NamedTuple of the fit results containing values and errors
+    * `report`: NamedTuple of the fit report which can be plotted
+"""
+function fit_lq_DEP(h::Histogram, ps::NamedTuple; uncertainty::Bool=true)
```
theHenks, line 13: Move the actual `Normal` fit maybe into another function. Best would be to include it in the `fit_single_trunc_gauss` function, to be able to also fit binned data. 
This would make the code more clear and readable.

src/lqcut.jl
```diff
@@ -0,0 +1,133 @@
...
+        @debug "n: $(v_ml.n)"
+
+        result = merge(v_ml, )
+    end
+    report = (
+            v = v_ml,
+            h = h,
+            f_fit = x -> Base.Fix2(f_lq_DEP, v_ml)(x),
+        )
+    return result, report
```
theHenks, line 87: The `result` later goes directly into the metadata, so here it is also nice to put the cut value in here. Also, where does the sideband substraction happen?

src/lqcut.jl
```diff
@@ -0,0 +1,133 @@
...
+end
+export fit_lq_DEP
+
+function lq_drift_time_correction(lq_DEP_dt, t_tcal)
```
theHenks, line 91: I don't understand what this is doing. Why does it return the `boxedges`?

theHenks (changes requested):
Rest of the comments are in the PR

oschulz:
@DaGeibl , can you re-create this PR, targeting the new dev branch?

CC @theHenks

PR #63, opened 2024-08-08
Peakshapes

LisaSchlueter:
- implementation of new peakshapes: high-energy tail `f_fit_tails`, exponential background `f_fit_bckExp`, and linear background `f_fit_bckSlope` 
- update of p-value calculation for ML fits: in `gof.jl` 
             - `p_value()` old default; based on least-squares
             - `p_value_poissonll()`, new default for calibration peak fits. based on poisson likelihood ratio
- adjustment of pseudo-prior
- remove LegendDataManagement dependency (and add copy of `calfunc.jl` in `/src` instead)
- update plot receipe for new peakshapes. plot all components


LisaSchlueter:
- [ ] parse `m_cal_simple` to `fit_single_peak_th228`; return µ, centroid and σ in ADC @LisaSchlueter 
- [ ] include centroid in fit result @LisaSchlueter 
- [ ] in data-flow: overwrite µ, centroid and σ with calibrated values @theHenks 
- [ ] in data-flow: create new energy-calibration processors with alternative fit function @theHenks (see Lisas branch in data flow)

PR #67, opened 2024-08-19
ctc plot recipe

LisaSchlueter:
Exchanged plot receipe for charge-trapping correction: Drift-time vs. energy now in 1 panel (overlay of before and after correction). 
**User needs to load StatsPlots (instead of Plots) for this recipe to work.**


theHenks (approved):

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -6,7 +6,7 @@ using RecipesBase
 using Unitful, Format, Measurements, LaTeXStrings
 using Measurements: value, uncertainty
 using StatsBase, LinearAlgebra
-
+using BAT
```
fhagemann, line 9: Could you replace the `BAT` dependency by a dependency on `KernelDensity`?
This would also require adding `KernelDensity` in the `Project.toml` with `[compat]` entry `"0.5, 0.6"`

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -446,72 +446,67 @@ end
...
-        ylabel := "Counts"
-        yscale := :log10
+        seriestype := :line
         subplot := 2
-        report_ctc.h_before
+        c := :binary
+        colorbar := :none
+        fill := true
+        label := "Before correction"
+        BAT.kde((ustrip(report_ctc.e_peak), report_ctc.qdrift_peak ./ maximum(report_ctc.qdrift_peak)))
```
fhagemann, line 495: The function `kde` is not defined in `BAT`, but in `KernelDensity`, which is exported into `BAT`.
It would be cleaner to remove the `BAT` dependency and replace it with `KernelDensity` for version control.

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -446,72 +446,67 @@ end
...
+        c := :plasma
+        colorbar := :none
+        fill := false
+        label := "After correction"
+        xlims := xl
+        ylims := (0, 1)
+        yformatter := :plain
+        xlabel := "Energy (keV)"
+        ylabel := "Drift time (a.u.)"
+        BAT.kde((ustrip(report_ctc.e_ctc), report_ctc.qdrift_peak ./ maximum(report_ctc.qdrift_peak)))
```
fhagemann, line 509: Also here

fhagemann (commented):
Could you also provide a preview of a plot produced with this plot recipe?

LisaSchlueter:
> Could you provide a preview of a plot produced with this plot recipe?
yes!
ctc_kde_p03_r000_V08682A

ext/LegendSpecFitsRecipesBaseExt.jl
LisaSchlueter: We need `BAT` anyway in `LegendSpecFits`. Isn't it more complicated to also add `KernelDensity`?

ext/LegendSpecFitsRecipesBaseExt.jl
oschulz: We'd like to get rid of KernelDensity in BAT, ideally ... :-)

ext/LegendSpecFitsRecipesBaseExt.jl
fhagemann: This would be the „clean“ way, to also have version control over KernelDensity.
As Oli pointed out, maybe BAT might get rid of that dependency in the near future, then it would be nice to have it separated here.

fhagemann:
The plot looks super nice!!
One thing: the drift times of millions of microseconds seem a bit long to me. Could you check the units?

theHenks:
I think here `QDrift` is plotted, which is the integrated drift time. So therefore the high numbers.

theHenks:
@LisaSchlueter Could you resolve the `KernelDensity` comments from Felix, then we can merge and use it in the next data processing.

LisaSchlueter:
- added `KernelDensity` 
- normalize drift time to 1. I did that to avoid confusion, because we actually plot the integrated drift time
- removed the previous ctc plot from the recipe

theHenks (commented):
Perfect, thanks @LisaSchlueter

PR #74, opened 2024-10-04
Lq updated

DaGeibl:
Things generally changed in this pull request:
- aoe_cut.jl: renamed lq_mode to inverted mode and updated doc string
- lqcut.jl : fixed bugs from issue,  `lq_drift_time_correction `now works with mode instead of detector type, changed fitting form binned to unbinned for lq box fit during drift time correction, added percentile mode for drift time, added propfunc, updated result&report and doc strings, updated `LQ_cut `after bug fix in fitting, renamed some variables
- singlefit: fixed bug that where the bins of the truncated hist used for the fit where moved one bin to the left, resulting in the function beeing shifted to the left
- test: wrote test for LQ_cut and lq_drift_time_correction
- plots recipe: added recipe functions to plot report of `lq_drift_time_correction `and `LQ_cut`

### DIFF
--- a/src/aoe_cut.jl
+++ b/src/aoe_cut.jl
@@ -134,8 +134,7 @@ Get the surrival fraction of a peak after a AoE cut value `aoe_cut` for a given 
 - `sf`: Surrival fraction
 - `err`: Uncertainties
 """
-function get_peak_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, peak::T, window::Vector{T}, aoe_cut::Unitful.RealOrRealQuantity,; 
-    uncertainty::Bool=true, lq_mode::Bool=false, bin_width_window::T=2.0u"keV", sigma_high_sided::Unitful.RealOrRealQuantity=Inf*unit(first(aoe)), fit_func::Symbol=:gamma_def) where T<:Unitful.Energy{<:Real}
+function get_peak_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, peak::T, window::Vector{T}, aoe_cut::Unitful.RealOrRealQuantity,; uncertainty::Bool=true, inverted_mode::Bool=false, low_e_tail::Bool=true, bin_width_window::T=2.0u"keV", sigma_high_sided::Unitful.RealOrRealQuantity=NaN) where T<:Unitful.Energy{<:Real}
     # estimate bin width
     bin_width = get_friedman_diaconis_bin_width(e[e .> peak - bin_width_window .&& e .< peak + bin_width_window])
     # get energy before cut and create histogram
@@ -145,13 +144,21 @@ function get_peak_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e
     # fit peak and return number of signal counts
     result_before, report_before = fit_single_peak_th228(peakhist, peakstats,; uncertainty=uncertainty, fit_func=fit_func)
 
-    # get energy after cuts
-    e_survived, e_cut = if !lq_mode
-        #normal aoe version
-        e[aoe_cut .< aoe .< sigma_high_sided], e[aoe .<= aoe_cut .|| aoe .>= sigma_high_sided]
+    # get e after cut
+    if !isnan(sigma_high_sided)
+        # TODO: decide how to deal with the high sided cut!
+        e = e[aoe .< sigma_high_sided]
+        aoe = aoe[aoe .< sigma_high_sided]
+    end
+
+    if inverted_mode == false
+        #normal aoe mode
+        e_survived = e[aoe_cut .<= aoe]
+        e_cut = e[aoe_cut .> aoe]
     else
-        #lq version
-        e[aoe .< aoe_cut .|| aoe .> sigma_high_sided], e[aoe_cut .<= aoe .<= sigma_high_sided]
+        #inverted mode for lq cut
+        e_survived = e[aoe_cut .>= aoe]
+        e_cut = e[aoe_cut .< aoe]
     end
     
     # estimate bin width
@@ -196,25 +203,17 @@ Get the surrival fraction of a continuum after a AoE cut value `aoe_cut` for a g
 - `n_after`: Number of counts after the cut
 - `sf`: Surrival fraction
 """
-function get_continuum_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, center::T, window::T, aoe_cut::Unitful.RealOrRealQuantity,; lq_mode::Bool=false, sigma_high_sided::Unitful.RealOrRealQuantity=Inf*unit(first(aoe))) where T<:Unitful.Energy{<:Real}
-    # scale unit
-    e_unit = u"keV"
-    # get energy around center
-    aoe = aoe[center - window .< e .< center + window]
-    e = e[center - window .< e .< center + window]
-    # get bin width
-    bin_width = get_friedman_diaconis_bin_width(e)
+function get_continuum_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, center::T, window::T, aoe_cut::Unitful.RealOrRealQuantity,; inverted_mode::Bool=false, sigma_high_sided::Unitful.RealOrRealQuantity=NaN) where T<:Unitful.Energy{<:Real}
     # get number of events in window before cut
-    n_before = length(e)
-    # get energy after cuts
-    e_survived, e_cut = if !lq_mode
-        #normal aoe version
-        e[aoe_cut .< aoe .< sigma_high_sided], e[aoe .<= aoe_cut .|| aoe .>= sigma_high_sided]
-    else
-        #lq version
-        e[aoe .< aoe_cut .|| aoe .> sigma_high_sided], e[aoe_cut .<= aoe .<= sigma_high_sided]
+    n_before = length(e[center - window .< e .< center + window])
+    # get number of events after cut
+    n_after = length(e[aoe .> aoe_cut .&& center - window .< e .< center + window])
+    if !isnan(sigma_high_sided)
+        n_after = length(e[aoe_cut .< aoe .< sigma_high_sided .&& center - window .< e .< center + window])
     end
-    n_after = length(e_survived)
+    if lq_mode == true
+        n_after = length(e[aoe .< aoe_cut .&& center - window .< e .< center + window])
+    end
 
     # calculate surrival fraction
     sf = n_after / n_before
--- a/src/aoe_cut.jl
+++ b/src/aoe_cut.jl
@@ -123,7 +123,7 @@ get_peaks_surrival_fractions(aoe, e, peaks, peak_names, left_window_sizes::Vecto
 
 
 """
-    get_peak_surrival_fraction(aoe::Array{T}, e::Array{T}, peak::T, window::Array{T}, aoe_cut::T,; uncertainty::Bool=true, low_e_tail::Bool=true) where T<:Real
+    get_peak_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, peak::T, window::Vector{T}, aoe_cut::Unitful.RealOrRealQuantity,; uncertainty::Bool=true, inverted_mode::Bool=false, low_e_tail::Bool=true, bin_width_window::T=2.0u"keV", sigma_high_sided::Unitful.RealOrRealQuantity=NaN) where T<:Unitful.Energy{<:Real}
 
 Get the surrival fraction of a peak after a AoE cut value `aoe_cut` for a given `peak` and `window` size while performing a peak fit with fixed position.
     
@@ -192,7 +192,7 @@ export get_peak_surrival_fraction
 
 
 """
-    get_continuum_surrival_fraction(aoe:::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, center::T, window::T, aoe_cut::Unitful.RealOrRealQuantity,; sigma_high_sided::Unitful.RealOrRealQuantity=NaN) where T<:Unitful.Energy{<:Real}
+    get_continuum_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, center::T, window::T, aoe_cut::Unitful.RealOrRealQuantity,; inverted_mode::Bool=false, sigma_high_sided::Unitful.RealOrRealQuantity=NaN) where T<:Unitful.Energy{<:Real}
 
 Get the surrival fraction of a continuum after a AoE cut value `aoe_cut` for a given `center` and `window` size.
 
@@ -211,7 +211,8 @@ function get_continuum_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantit
     if !isnan(sigma_high_sided)
         n_after = length(e[aoe_cut .< aoe .< sigma_high_sided .&& center - window .< e .< center + window])
     end
-    if lq_mode == true
+    #inverted mode for lq cut
+    if inverted_mode == true
         n_after = length(e[aoe .< aoe_cut .&& center - window .< e .< center + window])
     end
 
--- a/src/lqcut.jl
+++ b/src/lqcut.jl
@@ -1,5 +1,6 @@
 """
-    lq_drift_time_correction(lq_norm::Vector{Float64}, tdrift, e_cal; DEP_left=1589u"keV", det_icpc = false, DEP_right=1596u"keV", lower_exclusion=0.005, upper_exclusion=0.98, drift_cutoff_sigma=2.0)
+    lq_norm::Vector{Float64}, tdrift::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Array{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity;
+    DEP_edgesigma::Float64 = 3.0 , mode::Symbol = :gaussian, lower_exclusion::Float64 = 0.005, upper_exclusion::Float64 = 0.98, drift_cutoff_sigma::Float64 = 2.0, e_expression::Union{String,Symbol}="e")
 
 Perform the drift time correction on the LQ data using the DEP peak. The function cuts outliers in lq and drift time, then performs a linear fit on the remaining data. The data is Corrected by subtracting the linear fit from the lq data.
 # Returns
@@ -7,10 +8,9 @@ Perform the drift time correction on the LQ data using the DEP peak. The functio
     * `report`: NamedTuple of the histograms used for the fit
 """
 function lq_drift_time_correction(
-    lq_norm::Vector{Float64}, tdrift::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Array{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity;
+    lq_norm::Vector{Float64}, tdrift::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Vector{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity;
      DEP_edgesigma::Float64 = 3.0 , mode::Symbol = :gaussian, lower_exclusion::Float64 = 0.005, upper_exclusion::Float64 = 0.98, drift_cutoff_sigma::Float64 = 2.0,
-     e_expression::Union{String,Symbol}="e")
-
+     prehist_sigma::Float64 = 3.0, e_expression::Union{String,Symbol}="e")
 
     #calculate DEP edges
     DEP_left = DEP_µ - DEP_edgesigma * DEP_σ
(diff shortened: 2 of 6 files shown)


fhagemann:
I cleaned up this PR a bit and rebased it onto the current `dev` branch, so that one can more easily see, what exactly was changed in this PR.

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
+    lq_drift_time_correction(lq_norm::Vector{Float64}, tdrift, e_cal; DEP_left=1589u"keV", det_icpc = false, DEP_right=1596u"keV", lower_exclusion=0.005, upper_exclusion=0.98, drift_cutoff_sigma=2.0)
 
 Perform the drift time correction on the LQ data using the DEP peak. The function cuts outliers in lq and drift time, then performs a linear fit on the remaining data. The data is Corrected by subtracting the linear fit from the lq data.
 # Returns
     * `result`: NamedTuple of the corrected lq data, the box used for the linear fit and the drift time function
     * `report`: NamedTuple of the histograms used for the fit
 """
-function lq_drift_time_correction(lq_norm::Vector{Float64}, tdrift, e_cal, detectortype; DEP_left=1589u"keV", DEP_right=1596u"keV", lower_exclusion=0.005, upper_exclusion=0.98, drift_cutoff_sgima=2.0)
+function lq_drift_time_correction(
+    lq_norm::Vector{Float64}, tdrift::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Array{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity;
```
fhagemann, line 10: Any reason why you are sometimes using `Vector` and sometimes `Array`?

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
 
     #lq cutoff
-    #sort array to exclude outliers
-    sort_lq = sort(lq_DEP_dt)
-    low_cut = Int(round(length(sort_lq) * lower_exclusion)) # cut lower 0.5%
-    high_cut = Int(round(length(sort_lq) * upper_exclusion)) # cut upper 2%
-    lq_prehist = fit(Histogram, lq_DEP_dt, range(sort_lq[low_cut], sort_lq[high_cut], length=100))
+    #sort array to exclude outliers for better fit
+    low_cut_value = quantile(lq_DEP, lower_exclusion)  # 0.5% quantile
+    high_cut_value = quantile(lq_DEP, upper_exclusion)  # 2% quantile
```
fhagemann, line 26: Where is the array of LQ values sorted now?
Is some code missing or should this comment be deleted?

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
+
+    ideal_bin_width = LegendSpecFits.get_friedman_diaconis_bin_width(lq_DEP[low_cut_value .< lq_DEP .< high_cut_value])
```
fhagemann, line 28: You don't need to call `LegendSpecFits` within the package `LegendSpecFits`.

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
+    ideal_length = Int(round(abs(high_cut_value - low_cut_value) / ideal_bin_width))
+
+    lq_prehist = fit(Histogram, lq_DEP, range(low_cut_value, high_cut_value, length=ideal_length))
```
fhagemann, line 31: In `range`: you could use `step = ideal_bin_width`, or is there any particular reason why you are defining this via `length`?

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
     lq_prestats = estimate_single_peak_stats(lq_prehist)
+
     lq_start = lq_prestats.peak_pos - 3*lq_prestats.peak_sigma
     lq_stop = lq_prestats.peak_pos + 3*lq_prestats.peak_sigma
```
fhagemann, line 35: Is it intended that the `3` is hardcoded? Should this become a keyword argument of the function to be easily modified from outside? Or is that even `drift_cutoff_sigma`?

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
     #t_tcal cutoff; method dependant on detector type
-    if detectortype in [:ppc, :bege, :coax] 
-        drift_prehist = fit(Histogram, t_tcal, range(minimum(t_tcal), stop=maximum(t_tcal), length=100))
+    if mode == :gaussian 
+        
+        #ideal bin width for histogram
+        ideal_bin_width = LegendSpecFits.get_friedman_diaconis_bin_width(t_tcal)
+        ideal_length = Int(round((maximum(t_tcal) - minimum(t_tcal)) / ideal_bin_width))
+
+        drift_prehist = fit(Histogram, t_tcal, range(minimum(t_tcal), stop=maximum(t_tcal), length=ideal_length))
```
fhagemann, line 52: Again:
* remove `LegendSpecFits` (not needed)
* do we need `length = ideal_length` or could we use `step = ideal_bin_width` here?

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
         drift_prestats = estimate_single_peak_stats(drift_prehist)
         drift_start = drift_prestats.peak_pos - 3*drift_prestats.peak_sigma
         drift_stop = drift_prestats.peak_pos + 3*drift_prestats.peak_sigma
```
fhagemann, line 55: Again: `3` is hardcoded here.

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
...
         
-        drift_edges = range(drift_start, stop=drift_stop, length=71)
+        drift_edges = range(drift_start, stop=drift_stop, length=ideal_length)
         drift_hist_DEP = fit(Histogram, t_tcal, drift_edges)
         
-        drift_DEP_stats = estimate_single_peak_stats(drift_hist_DEP)
-        drift_result, drift_report = LegendSpecFits.fit_binned_gauss(drift_hist_DEP, drift_DEP_stats)
+        drift_result, drift_report = LegendSpecFits.fit_binned_trunc_gauss(drift_hist_DEP)
```
fhagemann, line 60: Remove `LegendSpecFits`

src/lqcut.jl
```diff
@@ -116,18 +137,19 @@ Evaluates the cutoff value for the LQ cut. The function performs a binned gaussi
     * `result`: NamedTuple of the cutoff value and the fit result
     * `report`: NamedTuple of the histograms used for the fit
 """
-function LQ_cut(DEP_µ, DEP_σ, e_cal, lq_classifier; lower_exclusion=0.005, upper_exclusion=0.95, cut_sigma=3.0)
+function LQ_cut(
+    DEP_µ::Unitful.Energy, DEP_σ::Unitful.Energy, e_cal::Vector{<:Unitful.Energy}, lq_classifier::Vector{Float64}; lower_exclusion::Float64=0.005, upper_exclusion::Float64=0.95, cut_sigma::Float64=3.0)
```
fhagemann, line 141: Do `DEP_µ` and `DEP_σ` always have energy units or could this also be just a `Real`?

src/lqcut.jl
```diff
@@ -116,18 +137,19 @@ Evaluates the cutoff value for the LQ cut. The function performs a binned gaussi
...
-    sort_lq = sort(lq_DEP)
-    low_cut = Int(round(length(sort_lq) * lower_exclusion)) # cut lower 0.5%
-    high_cut = Int(round(length(sort_lq) * upper_exclusion)) # cut upper 5%
-    prehist = fit(Histogram, lq_DEP, range(sort_lq[low_cut], sort_lq[high_cut], length=100))
+    low_cut_value = quantile(lq_DEP, lower_exclusion)  # 0.5% quantile
+    high_cut_value = quantile(lq_DEP, upper_exclusion)  # 95% quantile
+    prehist = fit(Histogram, lq_DEP, range(low_cut_value, high_cut_value, length=100))
     prestats = estimate_single_peak_stats(prehist)
     start = prestats.peak_pos - 3*prestats.peak_sigma
     stop = prestats.peak_pos + 3*prestats.peak_sigma
```
fhagemann, line 155: I see a lot of code duplication in `LQ_cut` and `lq_drift_time_correction`.
Could you export the identical parts into a separate function and call this helper function from `LQ_cut` and `lq_drift_time_correction`?

src/lqcut.jl
```diff
@@ -150,18 +172,18 @@ function LQ_cut(DEP_µ, DEP_σ, e_cal, lq_classifier; lower_exclusion=0.005, upp
...
         cut = cut_3σ,
-        fit_result = fit_result,
+
     )
 
     report = (
+        fit_result = fit_result,
         temp_hists = temp_hists,
         fit_report = fit_report,
     )
```
fhagemann, line 189: Is it correct that we don't want the `fit_result` in `result`?

src/lqcut.jl
```diff
@@ -71,36 +90,38 @@ function lq_drift_time_correction(lq_norm::Vector{Float64}, tdrift, e_cal, detec
...
     result_µ, report_µ = chi2fit(1, t_cut, lq_cut; uncertainty=true)
     parameters = mvalue(result_µ.par)
-    drift_time_func(x) = parameters[2] .* x .+ parameters[1] 
+    drift_time_func(x) = parameters[1] .+ parameters[2] .* x
 
-    #correct lq data with the linear fit to get lq classifier
-    lq_classifier = lq_norm .- drift_time_func(ustrip.(tdrift))
+    #property function for drift time correction
+    lq_class_func = "lq / $e_expression - ($(parameters[2]) * (qdrift / $e_expression) + $(parameters[1]))"
+    lq_class_func_generic = "lq / e  - (slope * tdrift + y_inter)"
```
fhagemann, line 110: `t_drift = q_drift / e` ?

fhagemann (changes requested):
Changes in this PR:
* in `get_peak_surrival_fraction`: renamed `lq_mode` to `inverted_mode`
* fix typo in `lq_drift_time_correction` keyword `drift_cutoff_sigma` (was `drift_cutoff_sgima` before)
* In `lq_drift_time_correction`: choose `t_lower` and `t_upper` for histogram generation based on `mode` (`:gaussian`, `:percentile`, `:double_gaussian`) rather than `detectortype` (`:ppc`, `:bege`, `:coax`, `:icpc`)
* Refactor `result` and `report` of both `lq_drift_time_correction` and `LQ_cut`.

src/lqcut.jl
```diff
@@ -1,58 +1,77 @@
 """
-    lq_drift_time_correction(lq_norm::Vector{Float64}, tdrift, e_cal; DEP_left=1589u"keV", det_icpc = false, DEP_right=1596u"keV", lower_exclusion=0.005, upper_exclusion=0.98, drift_cutoff_sgima=2.0)
+    lq_drift_time_correction(lq_norm::Vector{Float64}, tdrift, e_cal; DEP_left=1589u"keV", det_icpc = false, DEP_right=1596u"keV", lower_exclusion=0.005, upper_exclusion=0.98, drift_cutoff_sigma=2.0)
 
 Perform the drift time correction on the LQ data using the DEP peak. The function cuts outliers in lq and drift time, then performs a linear fit on the remaining data. The data is Corrected by subtracting the linear fit from the lq data.
 # Returns
     * `result`: NamedTuple of the corrected lq data, the box used for the linear fit and the drift time function
     * `report`: NamedTuple of the histograms used for the fit
 """
```
fhagemann, line 8: This docstring (and in general all docstrings of the functions you modified) are not up-to-date anymore.
(`mode` is missing here for example). Could you please always update the docstrings if you modify a function.
This would also make it way easier to check what exactly you modified and how it affects the function.

src/lqcut.jl
DaGeibl: No, what is type i should rather use?

src/lqcut.jl
DaGeibl: yes they have units

src/lqcut.jl
fhagemann: `Vector{T}` is just an alias for `Array{T,1}`. So, as long as `ecal` is always a 1-dimensional object (not a `Matrix`), I would avoid using `Array` and choose `Vector` (even better would be `AbstractVector` in case someone might pass something vector-like but not a `Vector`, e.g. an `AbstractRange`, `SubArray` or anything similar)

fhagemann:
Is there any small tests you could add to show that your method yields correct results? 
(some distribution with known sigma to check whether the LQ cut returns the correct three-sigma cut?)

src/lqcut.jl
```diff
@@ -52,7 +55,7 @@ function lq_drift_time_correction(
         drift_start = drift_prestats.peak_pos - prehist_sigma * drift_prestats.peak_sigma
         drift_stop = drift_prestats.peak_pos + prehist_sigma * drift_prestats.peak_sigma
         
-        drift_edges = range(drift_start, stop=drift_stop, length=ideal_length)
+        drift_edges = range(drift_start, stop=drift_stop, length=100)
```
fhagemann, line 58: Could you elaborate on choosing a fixed value of  `100` here?

src/lqcut.jl
```diff
@@ -12,6 +12,9 @@ function lq_drift_time_correction(
      DEP_edgesigma::Float64 = 3.0 , mode::Symbol = :gaussian, lower_exclusion::Float64 = 0.005, upper_exclusion::Float64 = 0.98, drift_cutoff_sigma::Float64 = 2.0,
      prehist_sigma::Float64 = 3.0, e_expression::Union{String,Symbol}="e")
 
+    #get energy units to remve later
+    e_unit = unit(first(e_cal))
```
fhagemann, line 16: This will fail if `e_cal` is empty (because then, there is not `first`). This could be avoided using `unit(eltype(e_cal))` if the type is defined correctly.

Are we sure that `e_cal` is never empty (or throw an error way earlier in the dataflow)?

DaGeibl: I copied what the "fit_aoe_corrections" in "aoe_fit_calibration.jl" where it is done similar. I can change it to unit(eltype(e_cal)), maybe then it also can be changed there? But i also think if the array is empty, the function will crash during the fitting.

src/lqcut.jl
DaGeibl: I get an error with the fit. I need to check in more detail what happens and why it happens. That didn't happen when i was fitting a whole partition(instead of like now only one run). So just a tempory fix till i (hopefully) fix it properly. The error is:

caused by: Error in drift time correction: BoundsError([20, 41, 27, 43, 42, 44, 22], (0:7,))
Stacktrace:
 [1] macro expansion
   @ /mnt/artemis02/users/gieb/MPP_Code/legend-julia-dataflow/processors/a_process_lq.jl:126 [inlined]
 [2] macro expansion
   @ ~/.julia/packages/ProgressMeter/kVZZH/src/ProgressMeter.jl:1008 [inlined]
 [3] top-level scope
   @ /mnt/artemis02/users/gieb/MPP_Code/legend-julia-dataflow/processors/a_process_lq.jl:113

caused by: BoundsError: attempt to access 7-element Vector{Int64} at index [0:7]
Stacktrace:
  [1] throw_boundserror(A::Vector{Int64}, I::Tuple{UnitRange{Int64}})
    @ Base ./abstractarray.jl:737
  [2] checkbounds
    @ ./abstractarray.jl:702 [inlined]
  [3] view
    @ ./subarray.jl:184 [inlined]
  [4] estimate_single_peak_stats_th228(h::Histogram{Int64, 1, Tuple{StepRangeLen{…}}})
    @ LegendSpecFits /mnt/artemis02/users/gieb/MPP_Code/SpecFits/src/peakstats.jl:59
  [5] fit_binned_trunc_gauss(h_nocut::Histogram{…}, cuts::@NamedTuple{…}; uncertainty::Bool)
    @ LegendSpecFits /mnt/artemis02/users/gieb/MPP_Code/SpecFits/src/singlefit.jl:360
  [6] fit_binned_trunc_gauss
    @ /mnt/artemis02/users/gieb/MPP_Code/SpecFits/src/singlefit.jl:351 [inlined]
  [7] fit_binned_trunc_gauss(h_nocut::Histogram{Int64, 1, Tuple{StepRangeLen{…}}})
    @ LegendSpecFits /mnt/artemis02/users/gieb/MPP_Code/SpecFits/src/singlefit.jl:351
  [8] lq_drift_time_correction(lq_norm::Vector{…}, dt_eff::Vector{…}, e_cal::Vector{…}, DEP_μ::Quantity{…}, DEP_σ::Quantity{…}; DEP_edgesigma::Float64, mode::Symbol, lower_exclusion::Float64, upper_exclusion::Float64, drift_cutoff_sigma::Float64, prehist_sigma::Float64, e_expression::String)
    @ LegendSpecFits /mnt/artemis02/users/gieb/MPP_Code/SpecFits/src/lqcut.jl:61
  [9] macro expansion
    @ /mnt/artemis02/users/gieb/MPP_Code/legend-julia-dataflow/processors/a_process_lq.jl:123 [inlined]
 [10] macro expansion
    @ ~/.julia/packages/ProgressMeter/kVZZH/src/ProgressMeter.jl:1008 [inlined]
 [11] top-level scope
    @ /mnt/artemis02/users/gieb/MPP_Code/legend-julia-dataflow/processors/a_process_lq.jl:113
Some type information was truncated. Use `show(err)` to see complete types.

fhagemann: The error message shows exactly where this comes from:

https://github.com/DaGeibl/LegendSpecFits.jl/blob/954d96aad356628db8da468bc748541bb24997b6/src/peakstats.jl#L58

as there is no entry more than 15 sigma away from the center position, the index defaults to length(array)-7. as your array only has 7 entries, it tries to access the array at index 0, which errors.

src/lqcut.jl
fhagemann: As long as we have some checks that ensure that the array is not empty, this can stay.

theHenks:
@DaGeibl  Looks good to me, but the tests need to be fixed. But it might be just a numerical issue, so maybe we fix this in a separate PR.
Also, could you add some more information in the PR desciption what exactly was changed now including a preview of your recipe? That helps for the documentation of the PRs.

DaGeibl:
Preview of the plots recipe added by this pull request:
report of lq_drift_time_correctio:
chose with Symbol plot_style between two versions. Version  :DEP where DEP events are shown with box highlighting events used for linear fit, together with the fit 
drift_time_vs_lq_plot_DEP_V08682A
Version :whole, not only DEP but all events.
drift_time_vs_lq_plot_whole_V08682A

report of LQ_cut:
Chose with Symbol plot_style between three versions. 
Version :lq_cut shows the lq vs energy 2dHistogramm together with line indicating lq cut value
lq_cut_plot_V08682A

Version :energyhist shows enrgy histogramm of 1)all events 2)events cu by lq and 3)events surviving lq cut
energy_hist_plot_V08682A

Version :cut_fraction shows the fraction of events cutted by LQ for different energies
cut_fraction_plot_V08682A

DaGeibl:
I also had a look at the test.

I checked the `fit_half_trunc_gauss` and did not find any bugs, for me the function does what its supposed to.
Only thing i thought about was that the bin width is calculated for the whole data, then the hist is truncated and for the truncated data the same bin width is used. I checked if there is an improvement calculating the bin width after truncating the data, which gives a smaller bin width, but the test still failed.
Also which of the 4 test for this function fails changes, it's not always the right or the left gauss.
So my best guess is that it's numerical instability. I also think if the other functions fit the whole gauss and this function only one side, its expected to do a "worse" fit right?
So maybe increase the allowed error tollerance for the fit_half_trunc_gauss? 

I also noticed my lq test sometimes(rarely) fails. I also think there could be a numerical issue, cause i am only fitting a gauß with 5000 events, because the detectors have events in that order. So maybe i could increase there the number of events or increase the allowed error tollerance slightly

DaGeibl:
I just found out that for the Random package there exist a "seed" that ensures that the random generated values are always the same. Should we maybe use that?

theHenks:
Thanks for checking the tests. I think the problem is the `approx` tests since we require only 5% of a difference. Maybe we just have to loosen that. I would go ahead and merge this PR and then open another issue for the tests in general. 
Mayabe also @fhagemann has a comment on that?

fhagemann:
If the tests usually pass when you re-run them, then it might make sense the loosen the allowed tolerance to isapprox.

src/lqcut.jl
```diff
@@ -1,60 +1,76 @@
 """
-    lq_drift_time_correction(lq_norm::Vector{Float64}, tdrift, e_cal; DEP_left=1589u"keV", det_icpc = false, DEP_right=1596u"keV", lower_exclusion=0.005, upper_exclusion=0.98, drift_cutoff_sgima=2.0)
+    lq_norm::Vector{Float64}, dt_eff::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Vector{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity;
```
theHenks, line 2: Is there a reason why we require here `Float64`? I find this a bit strict. Let's say at some point we wanna store DSP output as `Float32` to save some diskspace, all the routines would fail.
Let's change that.

src/aoe_cut.jl
```diff
@@ -149,7 +149,7 @@ get_peaks_surrival_fractions(aoe, e, peaks, peak_names, left_window_sizes::Vecto
 
 
 """
-    get_peak_surrival_fraction(aoe::Array{T}, e::Array{T}, peak::T, window::Array{T}, aoe_cut::T,; uncertainty::Bool=true, low_e_tail::Bool=true) where T<:Real
+    get_peak_surrival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, peak::T, window::Vector{T}, aoe_cut::Unitful.RealOrRealQuantity,; uncertainty::Bool=true, inverted_mode::Bool=false, low_e_tail::Bool=true, bin_width_window::T=2.0u"keV", sigma_high_sided::Unitful.RealOrRealQuantity=NaN) where T<:Unitful.Energy{<:Real}
```
theHenks, line 152: I don't understand this change. This is already changed in `dev` . This looks like a rebase problem.

src/lqcut.jl
```diff
@@ -1,60 +1,76 @@
...
-        t_lower = µ_t - drift_cutoff_sgima * σ_t
-        t_upper = µ_t + drift_cutoff_sgima * σ_t
-    elseif detectortype == :icpc
+        t_lower = µ_t - drift_cutoff_sigma * σ_t
+        t_upper = µ_t + drift_cutoff_sigma * σ_t
+
+    elseif mode == :double_gaussian # can be used for detectors with double peaks; not optimized yet
         #create histogram for drift time
-        drift_prehist = fit(Histogram, t_tcal, range(minimum(t_tcal), stop=maximum(t_tcal), length=100))
+        drift_prehist = fit(Histogram, dt_eff_DEP, range(minimum(dt_eff_DEP), stop=maximum(dt_eff_DEP), length=100))
```
theHenks, line 73: Is there no way around this 100 here? I can already see this failing...

theHenks (changes requested):
I added three comments. Mostly, my concerns here are regarding the dataflow. As a single routine this looks good!

theHenks:
@DaGeibl There are two requested changes open. My current ones and an old one from Felix. Could you resolve that and we merge, so that we can discuss this on Wednesday?

DaGeibl:
I fixed the first and the last comment, and I tried to rebase my branch but i don't really know if i was succesfull, If not I am not 100% sure what i am doing wrong. Maybe i could create a new branch and copy my lq files to the new branch then it would be cleaner, because with the rebase i think i added a lt duplicate commits(?)
Felix comments should all be fixed

fhagemann:
Sorry, but this commit history needs a fix, there are some many duplicates that should be avoided. 
Are you available tomorrow?
image

fhagemann:
Some of the tests are still failing. Is this unrelated?

DaGeibl:
the failling of the test is what florian mentioned earlier, that is nothing new

DaGeibl:
or can i in an other way fix the history?

fhagemann:
Usually, "clean" rebases should not create merge commits.. This here has quite a number of merge commits. If you have merge conflicts, they should be fixed but not by introducing a merge commit. Should we go through this tomorrow and fix this together?

fhagemann:
I force pushed again so fix and rebase the commit history

fhagemann (approved):
I would like to look at this in a bit more detail, but my changes were incorporated

theHenks:
Is this done and ready?

PR #77, opened 2024-10-09
Implementation of A/E fit background function with two EMGs 

verenaaur:
changed so far:
- changed markersize in plot recipe
- implemented option to choose between aoe fit background function with one EMG (default) or two EMGs (-> two fit_func 'aoe_one_bck' and 'aoe_two_bck')
- updated plot recipe: additional keyword "x_label" which is set to "Energy (keV)" by default; show A/E labels by using "x_label = "A/E"" like this:
`p = plot(report, x_label="A/E")`
- updated A/E pseudo priors (still some LogUniform distributions, though)
- updated A/E fit function (two EMGs) in peakshapes.jl file to include sigma = 0.0 case

current code to obtain Compton band histogram (when 'filtered_aoe' is the array with all the filtered aoe values of the CBs):
```
binsize = 0.05
h = fit(Histogram, filtered_aoe, -20:binsize:3)
ps = LegendSpecFits.estimate_single_peak_stats_psd(h)
result, report = LegendSpecFits.fit_single_aoe_compton(h, ps, fit_func=:aoe_two_bck, uncertainty=true) # use background function with two EMGs
p = plot(report, x_label="A/E") # use A/E labels instead of energy labels
```
(binsize 0.05 and interval [-20,3] is variable)

### DIFF
--- a/src/peakshapes.jl
+++ b/src/peakshapes.jl
@@ -260,12 +260,7 @@ function two_emg_aoe_compton_peakshape( # total fit function with two EMGs
     μ::Real, σ::Real, n::Real, background::Real, δ::Real, 
     μ2::Real, σ2::Real, background2::Real, δ2::Real
 )
-    # Check for sigma values being zero
-    if σ == 0.0 || σ2 == 0.0
-        @warn "Sigma value is zero which may cause numerical instabilities."
-        return 0.0 # return 0.0 or NaN
-    end
-    return n * gauss_pdf(x, μ, σ) + background * ex_gauss_pdf(-x, -μ, σ, δ) + background2 * ex_gauss_pdf(-x, -μ2, σ2, δ2)
+    return iszero(σ) || iszero(σ2) ? zero(x) : n * gauss_pdf(x, μ, σ) + background * ex_gauss_pdf(-x, -μ, σ, δ) + background2 * ex_gauss_pdf(-x, -μ2, σ2, δ2)
 end
 export two_emg_aoe_compton_peakshape
 


src/aoefit.jl
```diff
@@ -160,7 +160,7 @@ Perform a fit of the peakshape to the data in `h` using the initial values in `p
     * `result`: NamedTuple of the fit results containing values and errors
     * `report`: NamedTuple of the fit report which can be plotted
 """
-function fit_single_aoe_compton(h::Histogram, ps::NamedTuple; uncertainty::Bool=true, pseudo_prior::NamedTupleDist=NamedTupleDist(empty = true), fit_func::Symbol=:f_fit, background_center::Union{Real,Nothing} = ps.peak_pos, fixed_position::Bool=false)
+function fit_single_aoe_compton(h::Histogram, ps::NamedTuple; uncertainty::Bool=true, pseudo_prior::NamedTupleDist=NamedTupleDist(empty = true), fit_func::Symbol=:aoe_one_bck, background_center::Union{Real,Nothing} = ps.peak_pos, fixed_position::Bool=false)
```
theHenks, line 163: I think there is also the function `fit_aoe_compton` which would also need the the `fit_func` as a kwarg. Could you add that real quick? The default can just be a fill array similar to `fit_peaks` in for the energy fits.

src/peakshapes.jl
```diff
@@ -250,3 +250,25 @@ function double_gaussian(
...
+###############################################################
+# additional peakshape functions for two EMG background
+###############################################################
+
+function two_emg_aoe_compton_peakshape( # total fit function with two EMGs
+    x::Real, 
+    μ::Real, σ::Real, n::Real, background::Real, δ::Real, 
+    μ2::Real, σ2::Real, background2::Real, δ2::Real
+)
+    return n * gauss_pdf(x, μ, σ) + background * ex_gauss_pdf(-x, -μ, σ, δ) + background2 * ex_gauss_pdf(-x, -μ2, σ2, δ2)
```
theHenks, line 263: Let's catch also here the case for `sigma` = 0.0 similar to the energy peakshape. Otherwise we might run again into numerical instabilites.

theHenks (changes requested):
Some minor changes, then we can merge.

theHenks (approved):

PR #84, opened 2024-11-03
Remove `maxtime` kwargs when calling `Optimization.LBFGS`

fhagemann:
I spotted a small copy/paste typo in the kwargs handling of LBFGS-B in Optimzation:
When checking if `maxtime` is set, it returns that `abstol` is not used. I assume it should read `maxtime` is not used: https://github.com/SciML/Optimization.jl/blob/5cb17b6b61bedc92295dd8b2f607ab6fcbbc0db7/src/lbfgsb.jl#L47-L49
To omit the tons of warnings, I removed `maxtime` everywhere where `LBFGS` is used as solver (because it is not used anyhow).

### DIFF
--- a/src/memory_utils.jl
+++ b/src/memory_utils.jl
@@ -1,6 +1,6 @@
 
-optim_max_mem=10.0 # in GB for the whole current process ignoring if things could run in parallel
-optim_time_limit=20.0 # in seconds a single Optim.jl optimization step
+const optim_max_mem=10.0 # in GB for the whole current process ignoring if things could run in parallel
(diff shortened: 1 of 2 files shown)


theHenks:
As far as I understand we have then no `maxtime` set anymore. Is this just not possible for `LBFGS`?

fhagemann:
The current LBFGS implementation in `Optimization` does not use `maxtime`, see the source code here:
https://github.com/SciML/Optimization.jl/blob/master/src/lbfgsb.jl (The only occurrences being lines 39 and 47)

PR #85, opened 2024-11-03
Increase the number of samples in `singlefit` tests

### LINKED ISSUE #82: Check test values for `isapprox`
Some of our tests are failing due to too stringent requirements on the `issapprox` checks. 
We just study this real quick and update the values.

fhagemann:
This PR might close #82.

Fitting a half gauss to noisy data results in the tests being unreliable.
I increased the number of samples in the `singlefit` tests to decrease the variance in the fit results for `μ` and `σ`.

# Before this PR
The fit values for `μ` would easily be something like `0.06` or `-0.06`, such that the frequency with which they would exceed the `abstol` of `0.1` would be reached often enough for the tests to fail.
image
image

# After this PR
The variance of `μ` seems to be below `0.02`, such that the defined `abstol` of `0.1` is big enough to cover it.
However, it might still occur that the tests **rarely** fail
image
image

### DIFF
--- a/src/aoefit.jl
+++ b/src/aoefit.jl
@@ -180,7 +180,7 @@ function fit_single_aoe_compton(h::Histogram, ps::NamedTuple; uncertainty::Bool=
     # MLE
     optf = OptimizationFunction((u, p) -> ((-) ∘ f_loglike ∘ inverse(f_trafo))(u), AutoForwardDiff())
     optpro = OptimizationProblem(optf, v_init, [])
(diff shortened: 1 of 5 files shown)


fhagemann:
Question linked to this PR:
If the `half_trunc_gauss` results in this large variance for "low" statistics, how much can we trust the fit results if we apply this to low-statistic data? The left-sided trunc gauss shows a positive bias on `μ`, whereas the right-side trunc gauss shows a negative bias on `μ`.

fhagemann:
Some removals of `maxtime` in #84 were missing - I am adding them here in 570df70.

theHenks (approved):
❤️

PR #88, opened 2024-11-11
Sipm calibration routines

theHenks:
This PR includes a SiPM calibration routine fit based on the [`GaussianMixtures.jl`](https://github.com/davidavdav/GaussianMixtures.jl) package. 
- It fits a certain PE range with the mixture model and determines the PE positions including uncertainty handling based on ForwardDiff
- The `sipm_simple_calibration` function is adapted to have better output results
- Plot recipes for all outputs
The recipe can be controlled with the following `kwargs`:
- `xerrscaling`: scale all errors by a factor
- `show_residuals`: show residual panel
- `show_peaks`: show peak positions in plot
- `show_components`: show all mixtures components
image


fhagemann:
I was requested as reviewer.. should I still look through this or not?

PR #93, opened 2024-11-28
add ft_fwhm_tol to avoid artificially large ft in filter opt

LisaSchlueter:
I added a `ft_fwhm_tol ` parameter in the the flattop time optimization routines.  

Instead of just taking the minimum `fwhm`, we search for lowest "optimal" ft within `minimum(fwhm) + ft_fwhm_tol` to avoid artificially large ft.  The default tolerance is set to 0.1 keV.  We could put that in the metadata...

### DIFF
--- a/src/filter_optimization.jl
+++ b/src/filter_optimization.jl
@@ -106,7 +106,7 @@ export fit_fwhm_ft
 
 
 """
-    _fit_fwhm_ft(e_grid::Matrix, e_grid_ft::StepRangeLen{Quantity{<:T}, Base.TwicePrecision{Quantity{<:T}}, Base.TwicePrecision{Quantity{<:T}}, Int64}, rt::Unitful.RealOrRealQuantity, min_e::T, max_e::T, nbins::Int64, rel_cut_fit::T; default_ft::Quantity{T}=3.0u"µs") where {T <:Real}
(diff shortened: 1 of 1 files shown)


src/filter_optimization.jl
```diff
@@ -106,7 +106,7 @@ export fit_fwhm_ft
 
 
 """
-    _fit_fwhm_ft(e_grid::Matrix, e_grid_ft::StepRangeLen{Quantity{<:T}, Base.TwicePrecision{Quantity{<:T}}, Base.TwicePrecision{Quantity{<:T}}, Int64}, rt::Unitful.RealOrRealQuantity, min_e::T, max_e::T, nbins::Int64, rel_cut_fit::T; default_ft::Quantity{T}=3.0u"µs") where {T <:Real}
+    _fit_fwhm_ft(e_grid::Matrix, e_grid_ft::StepRangeLen{Quantity{<:T}, Base.TwicePrecision{Quantity{<:T}}, Base.TwicePrecision{Quantity{<:T}}, Int64}, rt::Unitful.RealOrRealQuantity, min_e::T, max_e::T, nbins::Int64, rel_cut_fit::T; default_ft::Quantity{T}=3.0u"µs", ft_fwhm_tol::Unitful.AbstractQuantity = 0.1u"keV") where {T <:Real}
```
theHenks, line 109: I would require that this parameter is a `Unitful.Energy{<:Real}` so that the units are already checked here. 
And maybe we could use something like `\Delta_fwhm_tol` or so? But naming is subjective and my naming is also not good, so no strong opinions here.

theHenks (changes requested):
Looks good to me, thanks @LisaSchlueter

src/filter_optimization.jl
theHenks: Done with latest commit

PR #94, opened 2024-12-04
Fixed bug: missing fit_func argument in aoe_cut.jl functions

DaGeibl:
I just noticed that with my LQ pull, there was a bug introduced, where the "fit_func" argument got lost from the get_peak_surrival_fraction. 
Probably happend during the rebasing with Felix when we tried to untangle my commit mess, where we instead of adding the changes of the new commit, kept an old commit resulting in a broken function.

I changed the function to what it was supposed to be with the discarded commit.

Currently due to this bug the A/E and LQ processors should be broken, which will be fixed by this.

### DIFF
--- a/src/aoe_cut.jl
+++ b/src/aoe_cut.jl
@@ -88,14 +88,14 @@ export get_low_aoe_cut
 
 
 """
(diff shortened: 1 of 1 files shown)


fhagemann:
Sorry! 🙏🏼

PR #102, opened 2024-12-11
adjust energy calibration to uncertainties in literature values

LisaSchlueter:
Reason for change:`th228_lines` will have uncertainty values after the upcoming change in legend-metadata. 

I changed the simple calibration so that it can handle these uncertainties. The other functions involved in the energy calibration worked as is.


fhagemann (approved):

fhagemann:
I pushed these changes onto `main` as well

PR #80, opened 2024-10-16
Updated docstrings

apearsonn:
Updated docstrings in the following format:
```julia
"""
function()
Description of function.
 Arguments
- 'x'
- 'y'
- 'z'

 Keywords
- 'x'

 Returns
- 'x'

"""
```
There are still functions that need arguments, keywords, returns or a description added. All of the missing descriptions are found in this Google Doc. 
https://docs.google.com/document/d/1-clJFUV2ZZ8kklChEZ_7bfOK89VDA-6_WGYy7SodkRs/edit?usp=sharing

PR #107, opened 2024-12-13
renamed variables in lq_cut, changed expression handling

DaGeibl:
Some changes for the lq_cut

- renamed some variables in lq_cut, and renamed the drift time correction to function `lq_ctc_correction`
- lq_cut now requires `lq_e_corr_expression `and `dt_eff_expression` to construct prop func
- added the option to use higher order corrections for the ctc fit
- use`cut_single_peak` insted of own pre-cutting method
- adjusted recipe to changes


theHenks:
All tests are failing. I think you need to rename things in there too

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -1390,7 +1390,7 @@ end
         # 2D histogram for LQ Cut
         xlabel := "Energy"
         ylabel := "LQ (A.U.)"
-        nbins := (0:6:3000, -2.0:0.012:4.0)
+        nbins := (0:6:3000, -0.1:0.0006:0.2)
```
theHenks, line 1393: Can you do the binning dynamically at least for the `LQ`? I encountered already several issues with extreme RAM usage during the dataflow when algorhtihms are binning "non-normal" data into extremely fine bins which allocates a lot of memory?

src/lqcut.jl
```diff
@@ -1,54 +1,46 @@
...
-    DEP_edgesigma::Float64=3.0 , mode::Symbol=:percentile, drift_cutoff_sigma::Float64 = 2.0, prehist_sigma::Float64=2.5, e_expression::Union{String,Symbol}="e", dt_eff_low_quantile::Float64=0.15, dt_eff_high_quantile::Float64=0.95)
+    lq_norm::Vector{<:AbstractFloat}, dt_eff::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Vector{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity; 
+    ctc_dep_edgesigma::Float64=3.0 , ctc_driftime_cutoff_method::Symbol=:percentile, lq_outlier_sigma::Float64 = 2.0, drift_time_outlier_sigma::Float64 = 2.0, prehist_sigma::Float64=2.5, lq_e_corr_expression::Union{String,Symbol}="(lq / e)", dt_eff_expression::Union{String,Symbol}="(qdrift / e)" ,ctc_dt_eff_low_quantile::Float64=0.15, ctc_dt_eff_high_quantile::Float64=0.95, pol_fit_order::Int=1) 
 
 Perform the drift time correction on the LQ data using the DEP peak. The function cuts outliers in lq and drift time, then performs a linear fit on the remaining data. The data is Corrected by subtracting the linear fit from the lq data.
```
theHenks, line 5: Could you add a description of the `kwargs`?

src/lqcut.jl
```diff
@@ -1,54 +1,46 @@
...
     lq_DEP = lq_norm[DEP_left .< e_cal .< DEP_right]
     dt_eff_DEP = ustrip.(dt_eff[DEP_left .< e_cal .< DEP_right])
 
-    # Calculate range to truncate lq data for outlier removal
-    ideal_bin_width = get_friedman_diaconis_bin_width(lq_DEP)
-    lq_prehist = fit(Histogram, lq_DEP, range(minimum(lq_DEP), maximum(lq_DEP), step=ideal_bin_width))
-    lq_prestats = estimate_single_peak_stats(lq_prehist)
-    lq_start = lq_prestats.peak_pos - prehist_sigma * lq_prestats.peak_sigma
-    lq_stop = lq_prestats.peak_pos + prehist_sigma * lq_prestats.peak_sigma
+    lq_precut = cut_single_peak(lq_DEP, minimum(lq_DEP), maximum(lq_DEP))
```
theHenks, line 22: The `cut_single_peak` function offers a couple `kwargs` which can be tuned. You might want to have a look into that if you want to parse them through here. 
Also, maybe you can find preset parameters for the big windowing rather than just `minimum(lq_DEP)` and `maximum(lq_DEP)`

src/lqcut.jl
```diff
@@ -1,54 +1,46 @@
...
 
     # truncated gaussian fit
-    lq_result, lq_report = fit_single_trunc_gauss(lq_DEP, (low=lq_start, high=lq_stop, max=NaN))
+    lq_result, lq_report = fit_single_trunc_gauss(lq_DEP, lq_precut)
```
theHenks, line 25: You can set `uncertainty=false` here to speed up the fit if you don't need the uncertainties.

src/lqcut.jl
```diff
@@ -65,10 +57,10 @@ function lq_drift_time_correction(
         σ_t = mvalue(drift_result.σ)
 
         #set cutoff in drift time dimension for later fit
-        t_lower = µ_t - drift_cutoff_sigma * σ_t
-        t_upper = µ_t + drift_cutoff_sigma * σ_t
+        t_lower = µ_t - drift_time_outlier_sigma * σ_t
+        t_upper = µ_t + drift_time_outlier_sigma * σ_t
 
-    elseif mode == :double_gaussian # can be used for detectors with double peaks; not optimized yet
+    elseif ctc_driftime_cutoff_method == :double_gaussian # can be used for detectors with double peaks; not optimized yet
```
theHenks, line 63: Maybe also consider here setting `uncertainty=false` if you are only interested in the Best fit values.

src/lqcut.jl
```diff
@@ -98,13 +90,13 @@ function lq_drift_time_correction(
...
     t_cut = dt_eff_DEP[lq_lower .< lq_DEP .< lq_upper .&& t_lower .< dt_eff_DEP .< t_upper]
 
-    #linear fit
-    result_µ, report_µ = chi2fit(1, t_cut, lq_cut; uncertainty=true)
-    parameters = mvalue(result_µ.par)
-    drift_time_func(x) = parameters[1] .+ parameters[2] .* x
+    #polynomial fit
+    result_µ, report_µ = chi2fit(pol_fit_order, t_cut, lq_cut; uncertainty=true)
+    par = mvalue(result_µ.par)
+    drift_time_func(x) =  sum((mvalue(par[i])) * x^(i-1) for i in eachindex(par))
```
theHenks, line 96: Is this supposed to just be the Polynomial? Because for that there is the `PolCalFunc` which you can use. In addition, I think the `chi2` report contains the function too.

src/lqcut.jl
```diff
@@ -1,54 +1,46 @@
 """
-    lq_norm::Vector{Float64}, dt_eff::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Vector{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity; 
-    DEP_edgesigma::Float64=3.0 , mode::Symbol=:percentile, drift_cutoff_sigma::Float64 = 2.0, prehist_sigma::Float64=2.5, e_expression::Union{String,Symbol}="e", dt_eff_low_quantile::Float64=0.15, dt_eff_high_quantile::Float64=0.95)
+    lq_norm::Vector{<:AbstractFloat}, dt_eff::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Vector{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity; 
+    ctc_dep_edgesigma::Float64=3.0 , ctc_driftime_cutoff_method::Symbol=:percentile, lq_outlier_sigma::Float64 = 2.0, drift_time_outlier_sigma::Float64 = 2.0, prehist_sigma::Float64=2.5, lq_e_corr_expression::Union{String,Symbol}="(lq / e)", dt_eff_expression::Union{String,Symbol}="(qdrift / e)" ,ctc_dt_eff_low_quantile::Float64=0.15, ctc_dt_eff_high_quantile::Float64=0.95, pol_fit_order::Int=1)
```
theHenks, line 3: Maybe we can stay with `qdrift` instead of `dt_eff` to avoid confusion with the other routines.

src/lqcut.jl
```diff
@@ -168,8 +159,10 @@ function LQ_cut(
...
     lq_prestats = estimate_single_peak_stats(hist_corrected)
-    lq_start = lq_prestats.peak_pos - truncation_sigma * lq_prestats.peak_sigma
-    lq_stop = lq_prestats.peak_pos + truncation_sigma * lq_prestats.peak_sigma
+    lq_start = lq_prestats.peak_pos - cut_truncation_sigma * lq_prestats.peak_sigma
+    lq_stop = lq_prestats.peak_pos + cut_truncation_sigma * lq_prestats.peak_sigma
+
+    #simplecuts benutzen
 
     # Fit the sideband subtracted histogram
     fit_result, fit_report = fit_binned_trunc_gauss(hist_corrected, (low=lq_start, high=lq_stop, max=NaN))
```
theHenks, line 168: I think, uncertainties here would be nice.

theHenks: I would also add more stuff into the result, in particular about this fit. Since otherwise, you cannot debug anything later.

theHenks (changes requested):
I have mainly some comments about the uncertainty handling, the returns and some of the naming.

ext/LegendSpecFitsRecipesBaseExt.jl
fhagemann: Well, it looks like here it is always the same "fineness" of binning and the only change is dividing the LQ values by 20 (why is that?)

src/lqcut.jl
fhagemann: A docstring should usually have the name of the function in it (in this case `lq_ctc_correction`).
Please add that to be consistent with other docstrings. Also have a look at #80

src/lqcut.jl
fhagemann: I agree

src/lqcut.jl
```diff
@@ -1,54 +1,46 @@
...
     * `report`: NamedTuple of the histograms used for the fit, the cutoff values and the DEP edges
 """
-function lq_drift_time_correction(
+function lq_ctc_correction(
     lq_norm::Vector{<:AbstractFloat}, dt_eff::Vector{<:Unitful.RealOrRealQuantity}, e_cal::Vector{<:Unitful.Energy{<:Real}}, DEP_µ::Unitful.AbstractQuantity, DEP_σ::Unitful.AbstractQuantity; 
-    DEP_edgesigma::Float64=3.0 , mode::Symbol=:percentile, drift_cutoff_sigma::Float64 = 2.0, prehist_sigma::Float64=2.5, e_expression::Union{String,Symbol}="e", dt_eff_low_quantile::Float64=0.15, dt_eff_high_quantile::Float64=0.95)
-
-    # get energy units to remve later
-    e_unit = unit(first(e_cal))
+    ctc_dep_edgesigma::Float64=3.0 , ctc_driftime_cutoff_method::Symbol=:percentile, lq_outlier_sigma::Float64 = 2.0, drift_time_outlier_sigma::Float64 = 2.0, prehist_sigma::Float64=2.5, lq_e_corr_expression::Union{String,Symbol}="(lq / e)", dt_eff_expression::Union{String,Symbol}="(qdrift / e)" ,ctc_dt_eff_low_quantile::Float64=0.15, ctc_dt_eff_high_quantile::Float64=0.95, pol_fit_order::Int=1)
```
fhagemann, line 12: why `ctc`?

src/lqcut.jl
```diff
@@ -127,23 +118,23 @@ function lq_drift_time_correction(
 
     return result, report
 end
-export lq_drift_time_correction
+export lq_ctc_correction
 
 """
-    DEP_µ::Unitful.Energy, DEP_σ::Unitful.Energy, e_cal::Vector{<:Unitful.Energy}, lq_classifier::Vector{Float64}; cut_sigma::Float64=3.0, truncation_sigma::Float64=2.0)
+    DEP_µ::Unitful.Energy, DEP_σ::Unitful.Energy, e_cal::Vector{<:Unitful.Energy}, lq_classifier::Vector{<:AbstractFloat}; cut_sigma::Float64=3.0, dep_sideband_sigma::Float64=4.5, cut_truncation_sigma::Float64=2.0)
```
fhagemann, line 124: Again here: the docstring is missing the name of the function: `lq_cut` and the description of the arguments (see #80)

src/lqcut.jl
```diff
@@ -127,23 +118,23 @@ function lq_drift_time_correction(
...
+function lq_cut(
+    DEP_µ::Unitful.Energy, DEP_σ::Unitful.Energy, e_cal::Vector{<:Unitful.Energy}, lq_classifier::Vector{<:AbstractFloat}; cut_sigma::Float64=3.0, dep_sideband_sigma::Float64=4.5, cut_truncation_sigma::Float64=2.0)
 
     # Define sidebands
-    lq_DEP = lq_classifier[DEP_µ - 4.5 * DEP_σ .< e_cal .< DEP_µ + 4.5 * DEP_σ]
-    lq_sb1 = lq_classifier[DEP_µ -  2 * 4.5 * DEP_σ .< e_cal .< DEP_µ - 4.5 * DEP_σ]
-    lq_sb2 = lq_classifier[DEP_µ + 4.5 * DEP_σ .< e_cal .< DEP_µ + 2 * 4.5 * DEP_σ]
+    lq_DEP = lq_classifier[DEP_µ - dep_sideband_sigma * DEP_σ .< e_cal .< DEP_µ + dep_sideband_sigma * DEP_σ]
+    lq_sb1 = lq_classifier[DEP_µ -  2 * dep_sideband_sigma * DEP_σ .< e_cal .< DEP_µ - dep_sideband_sigma * DEP_σ]
+    lq_sb2 = lq_classifier[DEP_µ + dep_sideband_sigma * DEP_σ .< e_cal .< DEP_µ + 2 * dep_sideband_sigma * DEP_σ]
```
fhagemann, line 137: Could you somehow make the casing (uppercase/lowercase) of `DEP`/`dep` consistent?

fhagemann (changes requested):

src/lqcut.jl
DaGeibl: The `fit_binned_trunc_gauss` returns values with uncertainties or do you mean something else?

src/lqcut.jl
DaGeibl: I copied the variable names from florians preliminary config (https://github.com/legend-exp/legend-jldataflow-config/blob/main/psd/p0%25_r0%25%25_psd.json). 
In general what pygama calls drift time correction we call carge trapping correction right?(e.g. for the A/E) I can also revert it and call everything drift time correction, dependant on what you find more understandable/ or consistent.

src/lqcut.jl
DaGeibl: Would you suggest in variables lower case for all variables, except the "peak_names"? Because there they usually appear uppercase

src/lqcut.jl
theHenks: Yes, but set the `uncertainty=true` kwarg explicitly.

DaGeibl: The fit is saved in the report. In gerneral what should be inside the result next to final values, like the cut, or the property function to get the classifier?

theHenks: Whatever you put into the result, will be later saved on disk and can be accessed for each run. This is then the only chance to do monitoring etc. So you should think about what you want to plot later and then fill the `result` accordingly.

src/lqcut.jl
theHenks: This is just a first proposal based on your old version. Please update the `jldataprod/config` accordingly with your new naming putting all your metadata pars in there and link the PR here.

src/lqcut.jl
theHenks: Whatever choice you take, it should be consistent.

src/lqcut.jl
DaGeibl: Ok, but then i would just use dt_eff to not confuse it with qdrift

src/lqcut.jl
DaGeibl: ok then I will write dep in variables lowercase, because all variables are typically lowercase,but when talking about it e.g in docstrings have the abbreviation uppercase

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -1275,7 +1275,7 @@ end
 
 # recipe for the lq_drift_time_correction report
 
-@recipe function f(report::NamedTuple{(:lq_prehist, :lq_report, :drift_prehist, :drift_report, :lq_box, :drift_time_func, :DEP_left, :DEP_right)}, e_cal, dt_eff, lq_e_corr, plot_type::Symbol)
+@recipe function f(report::NamedTuple{(:lq_report, :drift_report, :lq_box, :drift_time_func, :DEP_left, :DEP_right)}, e_cal, dt_eff, lq_e_corr, plot_type::Symbol)
```
theHenks, line 1278: What is the `plot_type` doing? Can this be a `kwarg`?

src/lqcut.jl
```diff
@@ -88,23 +92,23 @@ function lq_drift_time_correction(
         t_lower = minimum(x_at_threshold)
         t_upper = maximum(x_at_threshold)
     else
-        error("Mode $mode not supported")
+        error("Driftime cutoff method $ctc_driftime_cutoff_method not supported")
```
theHenks, line 95: If you want that the program exits here, you have to throw an `ArgumentError`

src/lqcut.jl
```diff
@@ -114,71 +118,86 @@ function lq_drift_time_correction(
     )
```
theHenks, line 118: There are still things in the `result` missing in my opinion. What is with the GOF, some debug parameters from the correction, the fit etc? 
I guess it is interesting to look later on at least at some variables and their behavior over time. Right now, nothing apart from the string will be saved.

src/lqcut.jl
```diff
@@ -114,71 +118,86 @@ function lq_drift_time_correction(
...
-    fit_result, fit_report = fit_binned_trunc_gauss(hist_corrected, (low=lq_start, high=lq_stop, max=NaN))
+    fit_result, fit_report = fit_binned_trunc_gauss(hist_corrected, (low=lq_start, high=lq_stop, max=NaN), uncertainty=true)
 
     #final cutoff value defined by "cut_sigma"
     cut_3σ = fit_result.μ + cut_sigma * fit_result.σ
 
     result = (
         cut = cut_3σ,
+        fit_µ = fit_result.μ,
+        fit_σ = fit_result.σ,
```
theHenks, line 200: Same comment here as above, at least information about the fit is necessary.

src/singlefit.jl
```diff
@@ -473,6 +473,7 @@ function fit_binned_trunc_gauss(h_nocut::Histogram, cuts::NamedTuple{(:low, :hig
     report = (
             f_fit = x -> Base.Fix2(f_fit, v_ml)(x) * bin_width,
             h = h_nocut,
+            h_cut = h,
```
theHenks, line 476: There is a recipe for this, did you change that too so that it still produces a plot?

theHenks (changes requested):
The `results` still need a fix.

ext/LegendSpecFitsRecipesBaseExt.jl
DaGeibl: `plot_type` specifies which one the different plots should be plottet. 
Should i implement this by picking one them as a default plot?

src/singlefit.jl
DaGeibl: I updated the recipe

src/aoe_cut.jl
```diff
@@ -199,25 +200,24 @@ Get the survival fraction of a continuum after a AoE cut value `aoe_cut` for a g
...
     e_survived, e_cut = if !inverted_mode
         #normal aoe version
         e[aoe_cut .< aoe .< sigma_high_sided], e[aoe .<= aoe_cut .|| aoe .>= sigma_high_sided]
     else
         #lq version
-        e[aoe .< aoe_cut .|| aoe .> sigma_high_sided], e[aoe_cut .<= aoe .<= sigma_high_sided]
+        e[aoe .< aoe_cut .&& .!aoe_lowcut_vec], e[aoe_cut .<= aoe .&& .!aoe_lowcut_vec]
     end
-    n_after = length(e_survived)
+    n_after = length(e_survived[center - window .< e_survived .< center + window])
```
fhagemann, line 220: How about something like
```
e_wdw_idx = findall(center - window .< e .< center + window)
```
and then index arrays with `e_wdw_idx` where needed?

DaGeibl:
I now also changed in function `aoe `to `psd_parameter `and `aoe_cut `to `psd_cut `to make clear it is also used for lq(and maybe future cuts) and not only aoe. if you want a shorter name than psd_parameter like `psd_par `or `psd_param `let me know then i change it

DaGeibl:
Compared to the last pull rewuest i removed the Union{Bitvector, nothing} type restriction and changed it to just Bitvector. In this version the `get_peaks_survival_fractions` needs two different versions, one with and one without Bitvector. I am not sure if this can be solved more elegeant as i did.

DaGeibl:
@theHenks @fhagemann This pull request is finished from my side, i would be happy if you could review it

src/aoe_cut.jl
```diff
@@ -242,5 +356,32 @@ function get_continuum_survival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantit
...
+
+    combined_cut_vector = if !inverted_mode
+        psd_cut .< psd_parameter .< sigma_high_sided .&& cut_vector
+    else
+        psd_parameter .< psd_cut .&& cut_vector
+    end
+
+    get_continuum_survival_fraction(combined_cut_vector, psd_parameter, e, center, window, psd_cut; 
+                                    sigma_high_sided=sigma_high_sided)
+end
```
fhagemann, line 384: Again, `cut_vector` could obtain a default value of `trues(length(psd_parameter))` to avoid having two functions.

src/aoe_cut.jl
```diff
@@ -88,14 +88,59 @@ export get_low_aoe_cut
...
+        v_result[i] = result_peak
+        v_report[i] = report_peak
+    end
+
+    # create result and report dict
+    result = Dict{Symbol, NamedTuple}(peak_names .=> v_result)
+    report = Dict{Symbol, NamedTuple}(peak_names .=> v_report)
+    
+    return result, report
+end
```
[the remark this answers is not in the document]
fhagemann: Just as noted below, `cut_vector` could for example receive a default value of `trues(length(psd_parameter))` to avoid having two methods that basically do the same thing (to avoid code duplication).

src/lqcut.jl
```diff
@@ -1,81 +1,88 @@
...
-    ideal_bin_width = get_friedman_diaconis_bin_width(lq_DEP)
-    lq_prehist = fit(Histogram, lq_DEP, range(minimum(lq_DEP), maximum(lq_DEP), step=ideal_bin_width))
-    lq_prestats = estimate_single_peak_stats(lq_prehist)
-    lq_start = lq_prestats.peak_pos - prehist_sigma * lq_prestats.peak_sigma
-    lq_stop = lq_prestats.peak_pos + prehist_sigma * lq_prestats.peak_sigma
+    #filter out NaN values
+    lq_dep_NaNfree = filter(isfinite, lq_dep)
+   
+    # precut lq data for fit
+    lq_precut = cut_single_peak(lq_dep_NaNfree, minimum(lq_dep_NaNfree), quantile(lq_dep_NaNfree, 0.99); relative_cut=ctc_lq_precut_relative_cut)
```
fhagemann, line 47: Should the `0.99` become a keyword argument, so that it can be changed by the user if wanted?

src/lqcut.jl
```diff
@@ -88,97 +95,111 @@ function lq_drift_time_correction(
         t_lower = minimum(x_at_threshold)
         t_upper = maximum(x_at_threshold)
     else
-        error("Mode $mode not supported")
+        throw(ArgumentError("Drift time cutoff method $ctc_driftime_cutoff_method not supported"))
     end
 
     #store cutoff values in box to return later    
     box = (lq_lower = lq_lower, lq_upper = lq_upper, t_lower = t_lower, t_upper = t_upper)
```
fhagemann, line 102: No need to change that here, but if you create a named tuple that has the field names identical to the variables that you're setting, you could write this in a more compact way:
```julia
box = (; lq_lower, lq_upper, t_lower, t_upper)
```

src/lqcut.jl
```diff
@@ -88,97 +95,111 @@ function lq_drift_time_correction(
...
-Evaluates the cutoff value for the LQ cut. The function performs a binned gaussian fit on the sidebandsubtracted LQ histogram and evaluates the cutoff value difined at 3σ of the fit.
 # Returns
     * `result`: NamedTuple of the cutoff value
     * `report`: NamedTuple of the fit result, fit report and temporary histograms
+
 """
-function LQ_cut(
-    DEP_µ::Unitful.Energy, DEP_σ::Unitful.Energy, e_cal::Vector{<:Unitful.Energy}, lq_classifier::Vector{<:AbstractFloat}; cut_sigma::Float64=3.0, truncation_sigma::Float64=2.0)
+function lq_cut(
+    dep_µ::Unitful.Energy, dep_σ::Unitful.Energy, e_cal::Vector{<:Unitful.Energy}, lq_classifier::Vector{<:AbstractFloat}; cut_sigma::Float64=3.0, dep_sideband_sigma::Float64=4.5, cut_truncation_sigma::Float64=3.5)
```
fhagemann, line 161: A quick comment why you changed the default value of the `cut_truncation_sigma` from `2.0` to `3.5`?

src/lqcut.jl
```diff
@@ -88,97 +95,111 @@ function lq_drift_time_correction(
...
+    weights_subtracted = hist_dep.weights .- hist_sb1.weights .- hist_sb2.weights
     hist_subtracted = Histogram(edges, weights_subtracted)
 
     # Replace negative bins with 0
     weights_corrected = max.(weights_subtracted, 0)
     hist_corrected = Histogram(edges, weights_corrected)
     
     # Create a named tuple of histograms for crosschecks
-    temp_hists = (hist_DEP = hist_DEP, hist_sb1 = hist_sb1, hist_sb2 = hist_sb2, hist_subtracted = hist_subtracted, hist_corrected = hist_corrected)
+    temp_hists = (hist_dep = hist_dep, hist_sb1 = hist_sb1, hist_sb2 = hist_sb2, hist_subtracted = hist_subtracted, hist_corrected = hist_corrected)
```
fhagemann, line 187: This notation could also be compactified

src/lqcut.jl
```diff
@@ -88,97 +95,111 @@ function lq_drift_time_correction(
...
     #get truncate values for fit; needed if outliers are present after in sideband subtracted histogram
     lq_prestats = estimate_single_peak_stats(hist_corrected)
-    lq_start = lq_prestats.peak_pos - truncation_sigma * lq_prestats.peak_sigma
-    lq_stop = lq_prestats.peak_pos + truncation_sigma * lq_prestats.peak_sigma
+    lq_start = lq_prestats.peak_pos - cut_truncation_sigma * lq_prestats.peak_sigma
+    lq_stop = lq_prestats.peak_pos + cut_truncation_sigma * lq_prestats.peak_sigma
 
     # Fit the sideband subtracted histogram
-    fit_result, fit_report = fit_binned_trunc_gauss(hist_corrected, (low=lq_start, high=lq_stop, max=NaN))
+    fit_result, fit_report = fit_binned_trunc_gauss(hist_corrected, (low=lq_start, high=lq_stop, max=NaN), uncertainty=true)
```
fhagemann, line 195: make `uncertainty` a keyword argument of `lq_cut` and pass it here?

test/test_lq.jl
```diff
@@ -29,12 +72,12 @@ using Distributions
...
+    result, report = lq_cut(dep_µ, dep_σ, e_cal, lq_classifier_combined)
 
-    # plot(report.temp_hists.hist_DEP, label="LQ SEP")
+    # plot(report.temp_hists.hist_dep, label="LQ SEP")
     # plot!(report.temp_hists.hist_sb1, label="LQ SB1")
     # plot!(report.temp_hists.hist_sb2, label="LQ SB2")
-    # plot!(report.temp_hists.hist_subtracted, label="DEP Subtracted")
+    # plot!(report.temp_hists.hist_subtracted, label="dep Subtracted")
     # plot(report.temp_hists.hist_corrected, label="original histogram")
     # plot!(report.fit_report.f_fit, label="Fit function")
```
fhagemann, line 82: Please remove from the tests what is not needed/used.

fhagemann (changes requested):
This PR features a lot of code duplication.
I commented where some functions can be combined to a single method by providing default values to the arguments.

Please also provide a plot using the new plot recipe, which makes it easier to review.

DaGeibl:
Here is an preview of the two plot recipes addedin this pull request.
Both plots are made using the report of LQ_cut. The first one is accesed by passing :sideband together with the report, lq_class and e_cal. It shows the sideband and the peak histogramms used for the sideband substraction. Here is an example:

grafik

The second plot is accesed by specifying :fit. 

grafik

src/aoe_cut.jl
```diff
@@ -88,14 +88,59 @@ export get_low_aoe_cut
...
+get_peaks_survival_fractions(psd_parameter, e, peaks, peak_names, left_window_sizes::Vector{<:Unitful.Energy{<:Real}}, right_window_sizes::Vector{<:Unitful.Energy{<:Real}}, psd_cut, cut_vector; kwargs...) = get_peaks_survival_fractions(psd_parameter, e, peaks, peak_names, [(l,r) for (l,r) in zip(left_window_sizes, right_window_sizes)], psd_cut, cut_vector; kwargs...)
+
+function get_peaks_survival_fractions(psd_parameter::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, peaks::Vector{<:T}, peak_names::Vector{Symbol}, windows::Vector{<:Tuple{T, T}}, psd_cut::Unitful.RealOrRealQuantity,; 
+    uncertainty::Bool=true, inverted_mode::Bool=false, bin_width_window::T=2.0u"keV", sigma_high_sided::Unitful.RealOrRealQuantity=Inf*unit(first(psd_parameter)), fit_funcs::Vector{Symbol}=fill(:gamma_def, length(peaks))) where T<:Unitful.Energy{<:Real}
```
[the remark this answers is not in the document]
DaGeibl: Yes excactly, to make clear, its not only aoe but all PSD cuts.

src/lqcut.jl
```diff
@@ -1,81 +1,88 @@
...
+    dt_eff_dep = ustrip.(dt_eff[dep_left .< e_cal .< dep_right])
 
-    # Calculate range to truncate lq data for outlier removal
-    ideal_bin_width = get_friedman_diaconis_bin_width(lq_DEP)
-    lq_prehist = fit(Histogram, lq_DEP, range(minimum(lq_DEP), maximum(lq_DEP), step=ideal_bin_width))
-    lq_prestats = estimate_single_peak_stats(lq_prehist)
-    lq_start = lq_prestats.peak_pos - prehist_sigma * lq_prestats.peak_sigma
-    lq_stop = lq_prestats.peak_pos + prehist_sigma * lq_prestats.peak_sigma
+    #filter out NaN values
+    lq_dep_NaNfree = filter(isfinite, lq_dep)
```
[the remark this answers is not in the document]
DaGeibl: I think the problem was that lq_cut and t_cut in line 105 and 106 are then not the same length

src/aoe_cut.jl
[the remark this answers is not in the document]
DaGeibl: Do I then even need two functions at all? can't i just have one function that takes psd_parameter, and an optional cut vector?

src/aoe_cut.jl
```diff
@@ -118,52 +163,62 @@ function get_peaks_survival_fractions(aoe::Vector{<:Unitful.RealOrRealQuantity},
...
-- `err`: Uncertainties
+A tuple containing:
+- `result`: NamedTuple with peak position, counts before and after the cut, survival fraction, and goodness-of-fit metrics.
+- `report`: NamedTuple with detailed fitting results for before and after the cut.
 """
-function get_peak_survival_fraction(aoe::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, peak::T, window::Vector{T}, aoe_cut::Unitful.RealOrRealQuantity,; 
-    uncertainty::Bool=true, inverted_mode::Bool=false, bin_width_window::T=2.0u"keV", sigma_high_sided::Unitful.RealOrRealQuantity=Inf*unit(first(aoe)), fit_func::Symbol=:gamma_def) where T<:Unitful.Energy{<:Real}  
-    # estimate bin width
+function get_peak_survival_fraction(cut_vector::BitVector, e::Vector{<:T}, peak::T, window::Vector{T}; 
+    uncertainty::Bool=true, bin_width_window::T=2.0u"keV", fit_func::Symbol=:gamma_def) where T<:Unitful.Energy{<:Real}
```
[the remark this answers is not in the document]
fhagemann: You right now have three methods for `get_peak_survival_fraction`.
I would rename this one to some internal function name (appending `_` up front)
and for the docstring, you could do something like
```julia
"""
get_peaks_survival_fractions(psd_parameter::Vector{<:Unitful.RealOrRealQuantity}, e::Vector{<:T}, peaks::Vector{<:T}, peak_names::Vector{Symbol}, windows::Vector{<:Tuple{T, T}}, psd_cut::Unitful.RealOrRealQuantity[, cut_vector::BitVector]; 
    uncertainty::Bool=true, inverted_mode::Bool=false, bin_width_window::T=2.0u"keV", sigma_high_sided::Unitful.RealOrRealQuantity=Inf*unit(first(psd_parameter)), fit_funcs::Vector{Symbol}=fill(:gamma_def, length(peaks))) where T<:Unitful.Energy{<:Real}
 
Get the survival fraction of a peak after a PSD cut value `psd_cut` for a given `peak` and `window` size while performing a peak fit with fixed position.
# Arguments
- `psd_parameter`: Vector of PSD parameter values.
- `e`: Vector of energy values.
- `peaks`: Vector of peak positions in energy.
- `peak_names`: Vector of symbols representing peak names.
- `windows`: Vector of tuples specifying the lower and upper energy bounds for fitting each peak.
- `psd_cut`: The PSD threshold used for classification.
- `cut_vector` (optional): BitVector indicating which events pass the PSD cut. If no `cut_vector` is given, no additional cut is applied.
# Keyword Arguments
- `uncertainty`: Whether to compute uncertainties (default: `true`).
- `inverted_mode`: If `true`, inverts the PSD cut logic (default: `false`).
- `bin_width_window`: Bin width for histogramming within peak windows (default: `2 keV`).
- `sigma_high_sided`: Upper bound for peak fitting beyond `psd_cut` (default: `Inf`).
- `fit_funcs`: Vector specifying the peak fitting function for each peak (default: `:gamma_def`).
"""
```

src/aoe_cut.jl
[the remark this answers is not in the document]
fhagemann: Nope, you would just need the one having the new implementation of `cut_vector`, with a default value if no value for `cut_vector` is passed.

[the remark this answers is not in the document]
DaGeibl: Ok to be clear instead of now 3 and I could go to 2 or 1? Because say i could just use 1 with the optional cut vector.

src/lqcut.jl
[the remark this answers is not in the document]
DaGeibl: yeah that would work. But what exactly makes this better that the original way?

src/aoe_cut.jl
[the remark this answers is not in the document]
fhagemann: You can also take the code from `get_peak_survival_fraction(cut_vector, ...)` and put it back into the original
`get_peak_survival_fraction(..., cut_vector = trues(length(psd_parameter)))` method.

src/aoe_cut.jl
fhagemann: Same things as discussed in https://github.com/legend-exp/LegendSpecFits.jl/pull/107#discussion_r1956241337

src/lqcut.jl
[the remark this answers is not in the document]
fhagemann: you avoid creating another array `lq_dep_NaNfree` which would cause additional allocations and use of memory

src/aoe_cut.jl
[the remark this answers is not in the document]
DaGeibl: What I was saying is, that i could could just merge all 3 `get_peak_survival_fraction` functions into a single one that has a `cut_vector = trues(length(psd_parameter)))` as argument. To not have the `_get_peak_survival_fraction function`, only a single one that does everything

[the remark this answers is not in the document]
fhagemann: That sounds good to me 👍

src/lqcut.jl
DaGeibl: i mean i could, but it's only there to remove extreme outliers so I don't i would ever modify that number

src/aoe_cut.jl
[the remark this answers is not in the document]
fhagemann: Let me know once this is pushed so I can review this PR in the final round.

fhagemann:
> The second plot is accesed by specifying :fit.
> 
> grafik

You could remove the xlabel from the upper plot and the title from the lower plot, and ideally make them touch.

DaGeibl:
@fhagemann i implemented the discussed changes to include a default cut vector of trues, to reduce the code duplication

fhagemann (approved):
@DaGeibl verified that the A/E output is not at all affected by the modifications in this PR, so this should be good to go from my side

theHenks (approved):

PR #108, opened 2024-12-16
Generic simple_calibration

LisaSchlueter:
- add a more generic `simple_calibration` function that can be accessed with keyword `calib_type == :gamma` which is useful for example for Co-60 data
- add a minimalistic gamma peak-shape, that contains only Gaussian signal and flat background to fit low-statistics data


src/simple_calibration.jl
```diff
@@ -81,7 +83,60 @@ function simple_calibration_th228(e_uncal::Vector{<:Real}, th228_lines::Vector{<
...
+    else
+        quantile(e_uncal, quantile_perc), length(gamma_lines)
+    end
+    @info "Identified most prominent peak (peak $(peak_idx), $(gamma_lines[peak_idx])). Peak guess: $peak_guess"
+    # get calibration constant for simple calibration
+    c = gamma_lines[peak_idx] / peak_guess
+    e_simple = e_uncal .* c
+    e_unit = u"keV"
+    # get peakhists and peakstats
+    peakhists, peakstats, h_calsimple, bin_widths = get_peakhists_th228(e_simple, gamma_lines, window_sizes; binning_peak_window=binning_peak_window, e_unit=e_unit)
```
fhagemann, line 117: `get_peakhists_th228` also works for generic `simple_calibration_gamma`?
Or would that also need a `get_peakhists_gamma`?

src/simple_calibration.jl
```diff
@@ -81,7 +83,60 @@ function simple_calibration_th228(e_uncal::Vector{<:Real}, th228_lines::Vector{<
...
 end
 
-
+"""
+    simple_calibration_gamma(e_uncal::Array, gamma_lines::Array, window_size::Float64=25.0, n_bins::Int=15000, quantile_perc::Float64=NaN, binning_peak_window::Float64=10.0, peak_quantile::ClosedInterval{<:Real} = 0.5..1.0)
+Perform a simple calibration for the uncalibrated energy array `e_uncal`
+* Find peaks within peak_quantile::ClosedInterval{<:Real} = 0.5..1.0 of the data
+* Estimate the calibration constant `c` by dividing the maximum gamma line energy by the peak guess
+* Return peakstats and peakhists of gamma lines for further calibration 
+"""
```
fhagemann, line 92: This docstring needs adjustment

src/simple_calibration.jl
```diff
@@ -81,7 +83,60 @@ function simple_calibration_th228(e_uncal::Vector{<:Real}, th228_lines::Vector{<
...
+        cts_peakpos = h_decon.weights[peakpos_idxs]
+        peakpos[argmax(cts_peakpos)], argmax(cts_peakpos)
+    else
+        quantile(e_uncal, quantile_perc), length(gamma_lines)
+    end
+    @info "Identified most prominent peak (peak $(peak_idx), $(gamma_lines[peak_idx])). Peak guess: $peak_guess"
+    # get calibration constant for simple calibration
+    c = gamma_lines[peak_idx] / peak_guess
+    e_simple = e_uncal .* c
+    e_unit = u"keV"
```
fhagemann, line 115: Should `e_unit` become a keyword argument, so you can pass it from the outside to the function?

src/specfit.jl
```diff
@@ -208,31 +212,34 @@ Get the FWHM of a peak from the fit parameters.
...
-        
-        tmp = x -> f_sigWithTail(x) - half_max_sig
-        roots_low = find_zero(tmp, e_low, maxiter=100)
-        roots_high = find_zero(tmp, e_high, maxiter=100)
-        return roots_high - roots_low
-    catch
-        return NaN
+    else
+        return 2 * sqrt(2 * log(2)) * v.σ
     end
```
fhagemann, line 243: This right now fails if `v` has no `σ`

fhagemann (commented):
Is this for Co60 or Co56?

src/simple_calibration.jl
LisaSchlueter: I wrote and tested this for my local Co60 data, but it should work for any other calibration spectra, because the anticipated gamma energies are passed as an argument. 
In principle we could consider re-writing/re-naming all `..._th228` calibration functions to be more generic for other gamma spectra. Some don't even need any modification at all and work out of the box for other sources. 
I didn't do that to keep the changes as minimal as possible.

LisaSchlueter:
> Is this for Co60 or Co56?

I wrote and tested this for my local Co60 data, but it should work for any other calibration spectra, because the anticipated gamma energies are passed as an argument.

src/specfit.jl
LisaSchlueter: yes, but there is no gamma fit function that doesn't have a `σ` as far as I know. I'll add a fall-back case anyway to be on the safe site.

src/simple_calibration.jl
LisaSchlueter: done

src/simple_calibration.jl
LisaSchlueter: I don't see why not. done

theHenks (changes requested):
In general, the PR looks good. Thanks @LisaSchlueter . 
My comments are mainly about the naming. 
Do we wanna stick to the isotope naming or start having more generic names?

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -490,6 +490,41 @@ end
     end
 end
 
+@recipe function f(report::NamedTuple{(:h_calsimple, :h_uncal, :c, :peak_guess, :peakhists, :peakstats)}; cal=true)
```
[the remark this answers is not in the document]
LisaSchlueter: This is _almost_ a duplicate of the existing recipe. However, the old one has some stuff hardcoded, e.g. "FEP" in the plot label. I didn't want to mess with the existing one, since we're int he middle of the processing.

src/simple_calibration.jl
```diff
@@ -18,23 +18,25 @@ Returns
...
     # remove calib type from kwargs
     @assert haskey(kwargs, :calib_type) "Calibration type not specified"
     calib_type = kwargs[:calib_type]
     # remove :calib_type from kwargs
     kwargs = pairs(NamedTuple(filter(k -> !(:calib_type in k), kwargs)))
     if calib_type == :th228
         @debug "Use simple calibration for Th228 lines"
-        return simple_calibration_th228(e_uncal, th228_lines, window_sizes,; kwargs...)
+        return simple_calibration_th228(e_uncal, gamma_lines, window_sizes,; kwargs...)
+    elseif calib_type == :gamma
```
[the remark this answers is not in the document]
LisaSchlueter: This works also for any calibration data. I think we should have only 1 generic `simple_calibration` function in the future and only special one if we can't make it work with the generic one

src/simple_calibration.jl
LisaSchlueter: I agree. I can take care of this beginning of next year. 

In that context, I'd suggest to change the julia energy-calibration metadata as well: 
1) Add a field "source" in the metadata, which is "th228" for LEGEND-200, but can also be something else. Like that the dataflow can pick the correct `gamma_lines` and `gamma_names` from the energy calibration config. 

OR

2) Rename "th228_lines" etc in the metadata to "gamma_lines". 

I prefer option 1, because this allows you to switch easily back and fourth between different calibration sources.

LisaSchlueter:
Thanks for your input! I really appreciate it. I suggest the following before merging this PR: 

**LegendSpecFits:**
- rename everything `th228` related into `gamma` 
- merge the two `simple_calibration` functions into one combined function. Same for plot recipe.
- test everything on th228 and co60 data

**julia-legend-dataflow:**
- adapt processor to new naming

**metadata / legend-jldataflow-config:**
- add `source` field (see above)

fhagemann:
That sounds like a great plan.
We can still define functions called `simple_calibration_th228` that will just call `simple_calibration_gamma`, something along
```julia
simple_calibration_th228(args...; kwargs...) = simple_calibration_gamma(args...; source = :th228, kwargs...)
```
where the keyword argument is `source`, `calib_type`, ... or whatever.

src/chi2fit.jl
```diff
@@ -94,7 +94,13 @@ function chi2fit(f_fit::Function, x::AbstractVector{<:Union{Real,Measurement{<:R
...
             chi2min = res.objective
             dof = length(x) - length(v_chi2)
-            pvalue = ccdf(Chisq(dof), chi2min)
+            if dof == 0
+                pvalue = NaN
+                  @warn "The number of fit parameters is equivalent to number of data points --> dof = 0 ; p-value not meaningful, set to NaN"
+            else 
+                pvalue = ccdf(Chisq(dof), chi2min)
+            end 
+
```
fhagemann, line 103:
```suggestion
            pvalue = if iszero(dof)
                @warn "The number of fit parameters is equivalent to number of data points --> dof = 0 ; p-value not meaningful, set to NaN"
                NaN
            else 
                ccdf(Chisq(dof), chi2min)
            end 
          
```

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -820,7 +855,8 @@ end
         if plot_ribbon
             ribbon := uncertainty.(report.f_fit.(0:1:1.2*value(maximum(report.x))))
         end
-        0:1:1.2*value(maximum(report.x)), value.(report.f_fit.(0:1:1.2*value(maximum(report.x))))
+        xstep = value(maximum(report.x))/100
```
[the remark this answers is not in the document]
LisaSchlueter: I tested it today on `228-Th` data and it worked well, same result.

src/simple_calibration.jl
```diff
@@ -18,23 +18,25 @@ Returns
...
+simple_calibration(e_uncal::Vector{<:Real}, gamma_lines::Vector{<:Unitful.Energy{<:Real}}, left_window_sizes::Vector{<:Unitful.Energy{<:Real}}, right_window_sizes::Vector{<:Unitful.Energy{<:Real}}; kwargs...) = simple_calibration(e_uncal, gamma_lines, [(l,r) for (l,r) in zip(left_window_sizes, right_window_sizes)],; kwargs...)
 
 
-function simple_calibration_th228(e_uncal::Vector{<:Real}, th228_lines::Vector{<:Unitful.Energy{<:Real}}, window_sizes::Vector{<:Tuple{Unitful.Energy{<:Real}, Unitful.Energy{<:Real}}},; n_bins::Int=15000, quantile_perc::Float64=NaN, binning_peak_window::Unitful.Energy{<:Real}=10.0u"keV")
+function simple_calibration_th228(e_uncal::Vector{<:Real}, th228_lines::Vector{<:Unitful.Energy{<:Real}}, window_sizes::Vector{<:Tuple{Unitful.Energy{<:Real}, Unitful.Energy{<:Real}}},; n_bins::Int=15000, quantile_perc::Float64=NaN, binning_peak_window::Unitful.Energy{<:Real}=10.0u"keV", e_unit::Unitful.EnergyUnits=u"keV")
```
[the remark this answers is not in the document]
LisaSchlueter: back to fixed unit

LisaSchlueter:
> In general, the PR looks good. Thanks @LisaSchlueter . My comments are mainly about the naming. Do we wanna stick to the isotope naming or start having more generic names?

We now have generic names `peak_search_gamma()`, `peakhists_gamma()`. There are still two options for `calib_type`
- `calib_type == :th228` 
- `calib_type == :gamma` 
In both cases, we apply the same methods to perform the simple calibration. The difference is that for th228, some values like the `peakfinder_threshold`, `peakfinder_σ`, `peak_quantile` and `bin_quantile` are harded-coded to good values for a LEGEND-like Th-228 spectrum.
In the case of `:gamma`, the user has more options to tune the simple calibration.

LisaSchlueter:
@theHenks Are you happy with your request?

LisaSchlueter:
Yes, let's do it after the CM then. I'll probably divide this into a few smaller PRs to make it easier to track the many small changes.

PR #111, opened 2024-12-19
Add Downgrade CI and update lower package compats

fhagemann:
The package versions in the `Project.toml` were a bit of a mess, so I cleaned them up.
To check that the packages still run with the lowest compats, I added the Downgrade CI on `1.10`, `1` and `pre`.

Changes made:
- The use of `Unitful.RealOrRealQuantity` requires `Unitful@1.6`
- The use of `value(::Quantity)` requires `Measurements@2.5`
- The use of `BAT` requires some updates listed in https://github.com/bat/BAT.jl/pull/466
- The use of `LBFGS` requires `Optimization@3.25`
  - `SciMLBase@2.30.0`
    - `SymbolicIndexingInterface@0.3.15`
      - `ArrayInterface@7.9`
        - `Adapt@4`
          - `StructArrays@0.6.17`
          - `FillArrays@1.4.1`
            - `ValueShapes@0.10.13`
            - `Distributions@0.25.87`
            - `Optim@1.7.5`
      
In addition, I included the bumps of `OptimizationOptimJL`, `OptimizationBBO` and `OptimizationNLopt` from #103, #104 and #105.


fhagemann:
This should be good to go.
I just wanna note that disentangling this was a freaking mess 😅 but this will probably keep nice and clean now! 😊

PR #112, opened 2025-02-03
Bug Fix `SiPM` Simple Calibration + Optimization

theHenks:
- Updated `sipm_simple_calibration` routines to be more robust for PE spectral detection by cutting out a possible noise peak followed by the use of `determine_calibration_constant_trough_peak_ratios` 
- Included `SiPM` optimization routines by sweeping through different filter length and the determination of resolution and gain which is used as objective. 
- Minor Bug Fixes


fhagemann:
This commit history needs some fixing.. There are some commits that are already on `main` in here..
Let me work on this.

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -509,14 +539,14 @@ end
         pps = report.peakpos
     end
     xlims := (0, last(first(h.edges)))
-    min_y = minimum(h.weights) == 0.0 ? 1e-3*maximum(h.weights) : 0.8*minimum(h.weights)
+    min_y = minimum(h.weights) == 0.0 ? 1e1 : 0.8*minimum(h.weights)
```
fhagemann, line 542: `1e1` is just `10`? :sweat_smile:

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -544,7 +574,7 @@ end
     ylabel := "Counts / $(round_wo_units(report_sipm.bin_width * 1e3, digits=2))E-3 P.E."
     xlims := (first(first(report_sipm.h_cal.edges)), last(first(report_sipm.h_cal.edges)))
     xticks := (ceil(first(first(report_sipm.h_cal.edges)))-0.5:0.5:last(first(report_sipm.h_cal.edges)))
-    min_y = minimum(report_sipm.h_cal.weights) == 0.0 ? 1e-3*maximum(report_sipm.h_cal.weights) : 0.8*minimum(report_sipm.h_cal.weights)
+    min_y = minimum(report_sipm.h_cal.weights) == 0.0 ? 1e1 : 0.8*minimum(report_sipm.h_cal.weights)
```
fhagemann, line 577: same here

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -154,6 +154,36 @@ end
...
+        [ustrip(Measurements.value(report.min_obj))]
+    end
+    @series begin
+        seriestype := :hspan
+        label := ""
+        color := :red
+        alpha := 0.1
+        ustrip.([Measurements.value(report.min_obj)-Measurements.uncertainty(report.min_obj), Measurements.value(report.min_obj)+Measurements.uncertainty(report.min_obj)])
+    end
+end
```
fhagemann, line 185: Could you provide an example plot how this looks like?
Makes it easier to judge than to browse through the code ;)

src/sipm_filter_optimization.jl
```diff
@@ -0,0 +1,139 @@
...
+"""
+function fit_sipm_wl(trig_max_grid::VectorOfVectors{<:Real}, e_grid_wl::StepRangeLen, thresholds::Vector{<:Real}=zeros(length(e_grid_wl));
+    min_pe_peak::Int=1, max_pe_peak::Int=5, n_fwhm_noise_cut::Real=2.0, peakfinder_threshold::Real=5.0, initial_max_amp::Real = 50.0, initial_max_bin_width_quantile::Real=0.9999, 
+    peakfinder_rtol::Real=0.1, peakfinder_α::Real=0.1, peakfinder_σ::Real=-1.0, 
+    min_pe_fit::Real=0.6, max_pe_fit::Real=3.5, Δpe_peak_assignment::Real=0.3)
+    
+    gain_wl    = Vector{Measurement{Float64}}(undef, length(e_grid_wl))
+    res_1pe_wl = Vector{Measurement{Float64}}(undef, length(e_grid_wl))
+    pos_1pe_wl = Vector{Measurement{Float64}}(undef, length(e_grid_wl))
+    success    = Vector{Bool}(zeros(length(e_grid_wl)))
```
fhagemann, line 23: Not sure if this makes a difference, but the convention would be to use
```julia
success = falses(length(e_grid_wl))
```

src/sipm_filter_optimization.jl
```diff
@@ -0,0 +1,139 @@
...
+            pos_1pe_wl[w] = first(result_fit.positions)
+            reports_simple[w] = report_simple
+            reports_fit[w] = report_fit
+            success[w] = true
+        catch e
+            @warn "Failed to process wl: $wl: $e"
+        end
+    end
+
+    thrs = if all(thresholds .== 0.0) ones(length(e_grid_wl)) else thresholds end
```
fhagemann, line 51: In case `thresholds` can have units, it would be safer to use
```julia
if all(iszero, thresholds) # ...
```

src/sipm_filter_optimization.jl
```diff
@@ -0,0 +1,139 @@
...
+        throw(ErrorException("No valid gain found, could not determine optimal window length"))
+    end
+    min_obj    = minimum(obj)
+    wl_min_obj = wls[findmin(obj)[2]]
+    min_res1pe = res_1pe_wl[success][findmin(obj)[2]]
+    min_gain   = gain_wl[success][findmin(obj)[2]]
+    min_pos1pe = pos_1pe_wl[success][findmin(obj)[2]]
+    min_threshold = thresholds[success][findmin(obj)[2]]
+    min_report_simple = reports_simple[success][findmin(obj)[2]]
+    min_report_fit = reports_fit[success][findmin(obj)[2]]
```
fhagemann, line 66: `findmin(obj)[2]` is just `argmin(obj)`.

Or you could change line 59 to `findmin` and apply it to lines 60-66:
```julia
min_obj, min_obj_idx = findmin(obj)
wl_min_obj = wls[min_obj_idx]
#...
```

src/sipm_filter_optimization.jl
```diff
@@ -0,0 +1,139 @@
...
+    
+    # fit histogram
+    result_trig, report_trig = if fit_thresholds
+        # generate cuts for thresholds
+        cuts_thres = cut_single_peak(thresholds, min_cut, max_cut; n_bins=n_bins, relative_cut=relative_cut)
+        # fit histogram
+        fit_binned_trunc_gauss(h, cuts_thres; uncertainty=uncertainty)
+    else
+        (μ = result_simple.μ_simple, σ = result_simple.σ_simple), h
+    end
```
fhagemann, line 134: In the `else` case: it is intended that the `report_trig` ends up being just a histogram?

src/sipm_simple_calibration.jl
```diff
@@ -27,34 +27,75 @@ function sipm_simple_calibration end
...
+)
+    
+    # Initial peak search
+    cuts_1pe = cut_single_peak(pe_uncal, initial_min_amp, initial_max_amp, relative_cut=relative_cut_noise_cut)
     
-    h_uncal, peakpos = find_peaks(pe_uncal; kwargs...)
+    bin_width_cut_min = cuts_1pe.max+n_fwhm_noise_cut*(cuts_1pe.high - cuts_1pe.max)
+    bin_width_cut = get_friedman_diaconis_bin_width(filter(in(bin_width_cut_min..quantile(pe_uncal, initial_max_bin_width_quantile)), pe_uncal))
+    peakpos = []
+    for bin_width_scale in 10 .^ range(0, stop=-3, length=50)
```
fhagemann, line 41: You could do:
```julia
for bin_width_scale in exp10.(range(0, stop=-3, length=50))
```

src/sipm_simple_calibration.jl
```diff
@@ -27,34 +27,75 @@ function sipm_simple_calibration end
...
+        @debug "Using bin width: $(bin_width_cut)"
+
+        bin_width_cut = bin_width_cut * bin_width_scale
```
fhagemann, line 44: I am not sure that this works as you intend it:
in every iteration of the `for`-loop, `bin_width` is multiplied with the value of the previous iteration.

MWE:
```julia
bin_width_cut = 1
for bin_width_scale in 1:10
    bin_width_cut = bin_width_cut * bin_width_scale
    print(bin_width_cut, ", ")
end
```
You probably want
```
1, 2, 3, 4, 5, 6, 7, 8, 9, 10,
```
but you get
```
1, 2, 6, 24, 120, 720, 5040, 40320, 362880, 3628800,
```

The solution would be to not call the variable you want to use `bin_width_cut` but something like `bin_width_cut_scaled`.

src/sipmfit.jl
```diff
@@ -104,10 +104,10 @@ function fit_sipm_spectrum(pe_cal::Vector{<:Real}, min_pe::Real=0.5, max_pe::Rea
 
     # get pe_pos
     get_pe_pos = pe -> let sel = in.(μ, (-Δpe_peak_assignment..Δpe_peak_assignment) .+ pe)
-        dot(view(μ,sel), view(w,sel)) / sum(view(w,sel))
+        dot(view(μ, sel), view(w, sel)) / sum(view(w, sel))
     end
     get_pe_res = pe -> let sel = in.(μ, (-Δpe_peak_assignment..Δpe_peak_assignment) .+ pe)
-        dot(view(σ,sel), view(w,sel)) / sum(view(w,sel))
+        sqrt(dot(view(σ, sel).^2, view(w, sel).^2))
     end
```
fhagemann, line 111: Just so that I understand: you add the resolutions in quadrature instead of just adding them all together?

fhagemann (changes requested):

src/sipm_filter_optimization.jl
theHenks: Since this is always based on `ADC` values and the type is `Real`, they can't have units

src/sipm_filter_optimization.jl
theHenks: Done

src/sipm_filter_optimization.jl
theHenks: Yes, to be able to plot it in both cases.

theHenks:
An example of the recipe plot for the filter optimization: 
l200-p03-r000-phy-S023-wl_sweep_sg

src/sipmfit.jl
theHenks: For a `GaussianMixtures`, the `σ` of a sum of normal distributions with weights is the formula shown. The former code was wrong.
The definition of resolution in this context can be discussed.

fhagemann:
The `Downgrade` tests seem to freeze because of `GaussianMixtures@0.3.11`
@theHenks: could you bump the version to `0.3.12`?

src/sipm_simple_calibration.jl
theHenks: Done, that was a mistake

ext/LegendSpecFitsRecipesBaseExt.jl
theHenks: s. PR

theHenks:
> The `Downgrade` tests seem to freeze because of `GaussianMixtures@0.3.11` @theHenks: could you bump the version to `0.3.12`?

Is this not already done on `main`?

src/sipm_simple_calibration.jl
```diff
@@ -38,13 +38,13 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
     bin_width_cut_min = cuts_1pe.max+n_fwhm_noise_cut*(cuts_1pe.high - cuts_1pe.max)
     bin_width_cut = get_friedman_diaconis_bin_width(filter(in(bin_width_cut_min..quantile(pe_uncal, initial_max_bin_width_quantile)), pe_uncal))
     peakpos = []
-    for bin_width_scale in 10 .^ range(0, stop=-3, length=50)
+    for bin_width_scale in exp10.(range(0, stop=-3, length=50))  # 1e-3 to 1
```
fhagemann, line 41: Just to make sure that this works as intended: The comment says `# 1e-3 to 1`, whereas the `for`-loop iterates from `1` to `1e-3` (reversed order). This is intended?

theHenks: According to CoPilot it is intended. I will remove th comment

fhagemann (approved):

PR #115, opened 2025-02-19
σ Plot Label Changes in Reports

verenaaur:
- Squares added to the labels in the σ plot (parameters of the fit are squared under the square root, which has been implemented in the code but is not yet shown in the labels) in the reports of the functions `fit_aoe_corrections` and `fit_aoe_compton_combined`

- Choose between labels where the first fit parameter is multiplied by e6 or by e3 (the label e3 is commented out)

### DIFF
--- a/src/LegendSpecFits.jl
+++ b/src/LegendSpecFits.jl
@@ -23,6 +23,7 @@ using GaussianMixtures
 using IntervalSets
 using InverseFunctions
 using IrrationalConstants
+using LaTeXStrings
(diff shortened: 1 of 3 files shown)


fhagemann:
I pushed a last commit to format the fit labels with `LaTeXStrings` which will allow to format the equation with an actual square root or a nicely formatted power of 10.

### Old recipe
image

### New recipe
image

fhagemann (approved):

PR #116, opened 2025-02-20
Remove filtering of `µ` and `σ` in the (individual) A/E correction fit

fhagemann:
In `fit_aoe_corrections`, the values of `µ` and `σ` were filtered by some sigma cut to remove outliers. However, this resulted in some "good" values for `µ` and `σ` at low energies to be removed.
I propose to remove this filter and accept the risk of individual outliers.

This only affects the individual A/E fits. The combined fit routine does not prefilter the values of `µ` and `σ`. This might also explain the observed differences between individual and combined fit (because the individual fits just fitted less data points).

### Before this PR
image

### After this PR
image (1)

Note: the nice fit labels are composed with the changes in #115, so I would also vote for merging both PRs at the same time.


theHenks (approved):
Thanks

theHenks:
CC @verenaaur

PR #117, opened 2025-02-20
Add tests for A/E correction fits (individual and combined)

fhagemann:
The last PRs #115 and #116 did not improve the code coverage.
This PR adds tests for the A/E energy-correction fit routines to compensate for that ;)

The fits are applied to some toy MC data that follows the expected energy dependence plus some smearing and exponential low A/E tail. The fit results agree with the inputs by 5% according to the tests, even though realistically, they agree within 1-2%. I am setting the threshold to 5% to minimize the risk of accidental test failures.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -9,5 +9,6 @@ Test.@testset "Package LegendSpecFits" begin
     include("test_singlefit.jl")
     include("test_docs.jl")
     include("test_lq.jl")
+    include("test_aoe.jl")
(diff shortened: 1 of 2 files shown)


fhagemann:
I increased the lower compat of Optimization to `3.26.0` because there seems to have been some problems with the return codes we use in the code in `3.25` that have been fixed in https://github.com/SciML/Optimization.jl/pull/766 and released in `3.26.0`. Downgrade tests should pass now.

PR #119, opened 2025-02-24
Avoid deprecated methods

fhagemann:
While running some tests, I got some warning about deprecated methods being used in LegendSpecFits, which we might wanna remove. This is done in this PR.

image


theHenks (approved):
Thanks!

PR #120, opened 2025-02-25
Patch `sipm_simple_calibration`

theHenks:
Updated `sipm_simple_calibration` for more stability in simple calibration routine.
- Bug Fix `peakfinder_σ` to also scale within each `bin_width` step
- Bug Fix `n_fwhm_noise_cut` for the `0.0` case
- Use `SavitzkyGolay` smoothing prior to peak search routine to smoothen `PE` distribution and enhance accuracy
- Inlcude failsafe routine by simple maximum detection based on smoothed histograms


src/sipm_simple_calibration.jl
```diff
@@ -35,33 +35,64 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
...
     peakpos = []
     for bin_width_scale in exp10.(range(0, stop=-3, length=50))
-        @debug "Using bin width: $(bin_width_cut)"
-
         bin_width_cut_scaled = bin_width_cut * bin_width_scale
+        @debug "Using bin width: $(bin_width_cut_scaled)"
         h_uncal_cut = fit(Histogram, pe_uncal, bin_width_cut_min:bin_width_cut_scaled:initial_max_amp)
-        if peakfinder_σ <= 0.0
-            peakfinder_σ = round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / 2.355)
+        peakfinder_σ_scaled = if peakfinder_σ <= 0.0
```
fhagemann, line 49: To avoid this (or have the error thrown somewhere else, you could explicitly set the type of `peakfinder_σ_scaled` to `Int` here:
```julia
peakfinder_σ_scaled::Int = if peakfinder_σ <= 0.0
    round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / 2.355)
else
    rount(Int, peakfinder_σ)
end

File: src/sipm_simple_calibration.jl [hunk 2]
```diff
@@ -35,33 +35,64 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
...
+            peakfinder_σ_scaled = round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / 2.355)
```
fhagemann, line 50: Small detail: the exact value would be `2 * sqrt(2 * log(2))` instead of `2.355`

src/sipm_simple_calibration.jl
```diff
@@ -35,33 +35,64 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
...
+        @warn "Failed to find peaks with peakfinder method, use alternative"
+        bin_width_cut_scaled = bin_width_cut * 0.5
+        @debug "Using bin width: $(bin_width_cut_scaled)"
+        
+        h_uncal_cut = fit(Histogram, pe_uncal, bin_width_cut_min:bin_width_cut_scaled:initial_max_amp)
+        peakfinder_σ_scaled = if peakfinder_σ <= 0.0
+            peakfinder_σ_scaled = round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / 2.355)
+        else
+            peakfinder_σ
+        end
```
fhagemann, line 82: Same thing here

src/sipm_simple_calibration.jl
```diff
@@ -35,33 +35,64 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
...
+        
+        # use SavitzkyGolay filter to smooth the histogram
+        sg_uncal_cut = savitzky_golay(h_uncal_cut.weights, ifelse(isodd(peakfinder_σ_scaled), peakfinder_σ_scaled, peakfinder_σ_scaled + 1), 3)
+        edges = (h_uncal_cut.edges[1][1:end-1] .+ h_uncal_cut.edges[1][2:end]) ./ 2
```
fhagemann, line 86:
```julia
edges = StatsBase.midpoints(first(h_uncal_cut.edges))
```

src/sipm_simple_calibration.jl
```diff
@@ -35,33 +35,64 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
...
+            if !isempty(peakpos) && length(peakpos) >= 2
                 break
             end
         end
     end
 
-    if isempty(peakpos) || length(peakpos) < 2
+    if length(peakpos) < 2
+        @warn "Failed to find peaks with peakfinder method, use alternative"
+        bin_width_cut_scaled = bin_width_cut * 0.5
```
fhagemann, line 74: Can you quickly comment on this: so basically you're forcing `bin_width_scale` to `0.5` and do the same thing again? Isn't `bin_width_scale = 0.5` already covered in the `for`-loop before (`10^-0.7`)?

fhagemann (commented):
There is some code duplication at the end of this function (the fallback method).
This is fine for me. If you want to remove it, you could export the duplicate code into a separate function that takes `bin_width_scale` as input and call it in the loop as well as in the fallback case.

Project.toml
```diff
@@ -34,6 +34,7 @@ RadiationSpectra = "4f207c7e-01da-51d7-a1a0-c8c06dd1d883"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 Roots = "f2b01f46-fcfa-551c-844a-d8ac1e96c665"
+SavitzkyGolay = "c4bf5708-b6a6-4fbe-bcd0-6850ed671584"
```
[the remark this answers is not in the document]
theHenks: I don't know why the compat was not updated automatically. Fixed now

src/sipm_simple_calibration.jl
theHenks: Done

src/sipm_simple_calibration.jl
theHenks: Done

src/sipm_simple_calibration.jl
```diff
@@ -47,7 +47,7 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
         @debug "Using bin width: $(bin_width_cut_scaled)"
         h_uncal_cut = fit(Histogram, pe_uncal, bin_width_cut_min:bin_width_cut_scaled:initial_max_amp)
         peakfinder_σ_scaled = if peakfinder_σ <= 0.0
-            peakfinder_σ_scaled = round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / 2.355)
+            peakfinder_σ_scaled = round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / 2 * sqrt(2 * log(2)))
```
fhagemann, line 50: Here you need brackets or the result will be wrong

fhagemann (dismissed):

src/sipm_simple_calibration.jl
theHenks: I want the `bin_width_cut_scaled` to be a small value for the fallback peakfinder, so it should just go smaller than the standard `bin_width` because then the smoothing is applied in the following lines so that too small `bin_width` effects will be smoothed out.
It also just worked for the current `SiPM` calibration runs, so it is just a guess.

src/sipm_simple_calibration.jl
theHenks: Scusa

src/sipm_simple_calibration.jl
theHenks: Done

src/sipm_simple_calibration.jl
fhagemann: 👍🏼

src/sipm_simple_calibration.jl
```diff
@@ -35,33 +35,64 @@ function sipm_simple_calibration(pe_uncal::Vector{<:Real};
...
-
         bin_width_cut_scaled = bin_width_cut * bin_width_scale
+        @debug "Using bin width: $(bin_width_cut_scaled)"
         h_uncal_cut = fit(Histogram, pe_uncal, bin_width_cut_min:bin_width_cut_scaled:initial_max_amp)
-        if peakfinder_σ <= 0.0
-            peakfinder_σ = round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / 2.355)
+        peakfinder_σ_scaled = if peakfinder_σ <= 0.0
+            round(Int, 2*(cuts_1pe.high - cuts_1pe.max) / bin_width_cut_scaled / (2 * sqrt(2 * log(2))) )
+        else
+            rount(Int, peakfinder_σ)
```
fhagemann, line 52: Not sure how this slipped through my review.. Sorry!

fhagemann: --> #131

PR #122, opened 2025-02-25
Omit performance warning when creating `OptimizationProblem`

fhagemann:
Tired of seeing this warning?
image

Say no more. We can replace `[]` by `()` (`Tuple` instead of `Vector`) to avoid this warning:
image

This is done everywhere in `LegendSpecFits` in this PR.


theHenks (approved):
Thanks, this also reduces the messages I see in the dataflow by a lot

PR #124, opened 2025-02-27
updated LQ sideband selection logic

DaGeibl:
This pull request changes how the LQ sideband are generated. Currently the peak is defined as 4.5 dep_sigma(value from calibration) around the literature value of the DEP. The sidebands lie between 4.5 and 9 sigma on both sides of the peak band. For Detectors with low resolution this can result in a case where the sidebands reach into the 212-Bi FEP. To avoid this, I moved the both sidebands to the energies below the DEP.

Plot of Detector with normal energy res:
l200-part01-cal-B00076C-energy_spectrum_lq_classifier

Plot of Detector with bad energy res before PR
l200-part01-cal-V08682A-energy_spectrum_lq_classifier

Plot of same Detector after PR
sideband

Also added with this PR is a recipe for the plot above, that can be implemented into the lq_processors

### DIFF
--- a/ext/LegendSpecFitsRecipesBaseExt.jl
+++ b/ext/LegendSpecFitsRecipesBaseExt.jl
@@ -1476,7 +1476,7 @@ end
 
 # recipe for the lq_cut report
 
-@recipe function f(report::NamedTuple{(:cut, :fit_result, :temp_hists, :fit_report)}, lq_class::Vector{Float64}, e_cal, plot_type::Symbol)
+@recipe function f(report::NamedTuple{(:cut, :fit_result, :temp_hists, :fit_report, :dep_σ, :edges)}, lq_class::Vector{Float64}, e_cal, plot_type::Symbol)
 
     # Extract cutvalue from the report
     cut_value = Measurements.value.(report.cut)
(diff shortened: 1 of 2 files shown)


ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -1662,10 +1662,40 @@ end
             report.temp_hists.hist_sb2
         end
 
-
+    elseif plot_type == :energy_spectrum
```
theHenks, line 1665: This "multiple-recipes-in-one" makes the code really hard to read, but this can stay for now since we anyhow move over to `LegendMakie.jl`

theHenks (approved):
Looks good. Did you test this on real data already?

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -1662,10 +1662,40 @@ end
...
+        # Plot energy spectrum with DEP and sideband regions
+        left_margin := -2Plots.mm
+        bottom_margin := -4Plots.mm
+        top_margin := -3Plots.mm
```
fhagemann, line 1669: Avoid using `Plots` in the extension.
You can rewrite `-2Plots.mm` as `(-2, :mm)` and so on.

fhagemann (changes requested):

ext/LegendSpecFitsRecipesBaseExt.jl
```diff
@@ -1662,10 +1662,40 @@ end
...
+        thickness_scaling := 1.6
+        size := (1200, 900)
+        framestyle := :box
+        formatter := :plain
+        xlabel := "Energy"
+        ylabel := "Counts"
+        
+        @series begin
+            seriestype := :stephist
+            label := "Energy Spectrum (σ: $(report.dep_σ))"
```
fhagemann, line 1679: Can you round the value of sigma to either 2 or 3 digits to avoid long labels?

src/lqcut.jl
```diff
@@ -159,11 +159,29 @@ function lq_cut(
...
+        lq_sb1 = lq_classifier[sb1_edge .< e_cal .< DEP_edge_left]
+        lq_sb2 = lq_classifier[DEP_edge_right .< e_cal .< sb2_edge]
+    else
+        sb1_edge = dep_µ - 2 * dep_sideband_sigma * dep_σ  
+        sb2_edge = dep_µ - 3 * dep_sideband_sigma * dep_σ
+
+        lq_dep = lq_classifier[DEP_edge_left .< e_cal .< DEP_edge_right]
+        lq_sb1 = lq_classifier[sb1_edge .< e_cal .< DEP_edge_left]
+        lq_sb2 = lq_classifier[sb2_edge .< e_cal .< sb1_edge]
+    end
```
fhagemann, line 180: Are there cases where `σ` ends up being so high, that it reaches into the peak to the left?

fhagemann (changes requested):

ext/LegendSpecFitsRecipesBaseExt.jl
theHenks: Where ist the change commit @DaGeibl ?

ext/LegendSpecFitsRecipesBaseExt.jl
theHenks: Same here, I don't see it changed yet. Just the conversation was resolved.

DaGeibl:
@theHenks I think there is a GitHub issue, or I do sth Wrong. On my DaGeibl/Main Branch there is the pushed Committee That resolves Felix Comments. Of I Look at this PR i See That its still „Processing Updates“. Did you encounter sth Like That before?

fhagemann:
Some times a good old force push fixes things ^^

fhagemann (approved):

theHenks (approved):

PR #126, opened 2025-02-28
Added missing bracket

DaGeibl:
Hi, I noticed that something strange was happening with the LQ processor. I noticed that the dirft time correction property function got broken during Florians debugging. Previoues to this PR the `lq_ctc_correction`() function expected the `ctc_driftime_expression` to have brackets. I now changed this so that the function also returns the correct property function if `ctc_driftime_expression` does not have explicitly brackets.

@theHenks  I also noticed that during your debugging you encountered some problems with my plots. This happened because the the wrong ctc correction broke them. With this fix they all work again as intended and it would suggest to include the `lq_cut` plot again as it is currently commented out. Except you had another reason the remove it from the lq processors

### DIFF
--- a/src/lqcut.jl
+++ b/src/lqcut.jl
@@ -109,7 +109,7 @@ function lq_ctc_correction(
     pol_fit_func = report_µ.f_fit
 
     # property function for drift time correction
-    lq_class_func = "$lq_e_corr_expression - " * join(["$(par[i]) * $dt_eff_expression^$(i-1)" for i in eachindex(par)], " - ")
+    lq_class_func = "$lq_e_corr_expression - " * join(["($(par[i]) * $dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
     lq_class_func_generic = "lq / e  - (slope * qdrift / e + y_inter)"
 
     # create result and report
--- a/src/lqcut.jl
+++ b/src/lqcut.jl
@@ -109,7 +109,7 @@ function lq_ctc_correction(
     pol_fit_func = report_µ.f_fit
 
     # property function for drift time correction
-    lq_class_func = "$lq_e_corr_expression - " * join(["($(par[i]) * $dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
+    lq_class_func = "$lq_e_corr_expression - " * join(["$(par[i]) * ($dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
     lq_class_func_generic = "lq / e  - (slope * qdrift / e + y_inter)"
 
     # create result and report


fhagemann (approved):

theHenks:
> Hi, I noticed that something strange was happening with the LQ processor. I noticed that the dirft time correction property function got broken during Florians debugging. Previoues to this PR the `lq_ctc_correction`() function expected the `ctc_driftime_cutoff_method` to have brackets. I now changed this so that the function also returns the correct property function if `ctc_driftime_cutoff_method` does not have explicitly brackets.

How did the the property function break during my debugging? I never touched this code??
> 
> @theHenks I also noticed that during your debugging you encountered some problems with my plots. This happened because the the wrong ctc correction broke them. With this fix they all work again as intended and it would suggest to include the `lq_cut` plot again as it is currently commented out. Except you had another reason the remove it from the lq processors. 

The issue with the plotting is related to `NaN` values within your `Vector`s to be plotted, in particular the `histogram2d` plots. I don't see how this PR is fixing this?

fhagemann (dismissed):
Can you continue discussing the `NaN` plotting issue somewhere else, for example by opening a new issue?
I feel like this PR fixes a very obvious and straight-forward-to-fix bug and should be merged ASAP.

src/lqcut.jl
```diff
@@ -109,7 +109,7 @@ function lq_ctc_correction(
     pol_fit_func = report_µ.f_fit
 
     # property function for drift time correction
-    lq_class_func = "$lq_e_corr_expression - " * join(["$(par[i]) * $dt_eff_expression^$(i-1)" for i in eachindex(par)], " - ")
+    lq_class_func = "$lq_e_corr_expression - " * join(["($(par[i]) * $dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
```
fhagemann, line 112: Wait! Is this even correct? or should this be:
```julia
q_class_func = "$lq_e_corr_expression - " * join(["$(par[i]) * ($dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
```

fhagemann, line 112:
```suggestion
    lq_class_func = "$lq_e_corr_expression - " * join(["$(par[i]) * ($dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
```

DaGeibl: Good catch! No you are Right there I made a Mistake 😅
It should not be raised by the power

theHenks:
> Can you continue discussing the `NaN` plotting issue somewhere else, for example by opening a new issue? I feel like this PR fixes a very obvious and straight-forward-to-fix bug and should be merged ASAP.

I created a new issue

src/lqcut.jl
fhagemann: Done

theHenks:
I will go ahead and merge since I want to test this on data

PR #132, opened 2025-03-09
Replace all report `Dicts` by `OrderedDict`

fhagemann:
Addressing the issue https://github.com/legend-exp/LegendMakie.jl/issues/29#issue-2903483141:
> The peaks are always plotted in reverse order (from highest energy peak to lowest). Unsure how much work this is, but it would be cool if they were plotted from lowest to highest peak position.

This is due to the `report` being a `Dict` which is by default unordered.
How, such `reports` will be `OrderedDict` to preserve the order of the individual reports.

Note: this PR introduces a new light dependency of `OrderedCollections`.


theHenks (approved):
Perfect, thanks! This will make plotting much nicer. 
Did you test all functions on data? Otherwise this is for me ready to be merged

fhagemann:
No, I did not test this yet on data, but I will within the next hours.

theHenks:
I tested the energy part and verified that it works. Let's merge!

PR #134, opened 2025-03-13
Added Sort for peakpos based on ADC energy value

RyutaroMatsumoto:
In `simple_calibration.jl` added Sort for `peakpos` based on ADC energy value to avoid sorting malfunction, which sometimes happens and leads to wrong calibration results.

### DIFF
--- a/src/simple_calibration.jl
+++ b/src/simple_calibration.jl
@@ -138,9 +138,23 @@ function peak_search_gamma(e_uncal::Vector{<:Real}, gamma_lines::Vector{<:Unitfu
         quantile(e_uncal, quantile_perc), length(gamma_lines)
     end
 
-    @debug "Identified most prominent peak at $(round(peak_guess, digits = 2)) - literature value: $(gamma_lines[peak_idx])"
+    # 1) sort by ADC x value (accending)                 
+    sorted_peakpos     = peakpos[sortperm(peakpos)]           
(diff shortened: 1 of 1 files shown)


fhagemann (commented):
Thanks for this PR.
Some comments from my side:
- could you replace `@info` with `@debug` so that the logging is shown in debug mode only (or add a `verbose` keyword to omit the `@info` output)?
- is this the correct place to put it? I somehow feel like this should go directly into `RadiationSpectra` applied to the output of `peakfinder` before returning it (?). This way, everybody using `RadiationSpectra.peakfinder` could profit from this.


Looking at LEGEND analysis:
- @theHenks would this interfere with finding the Tl208 FEP vs. finding the pulser peak if we sort them by ADC?

LisaSchlueter:
> Thanks for this PR. Some comments from my side:
> 
> * could you replace `@info` with `@debug` so that the logging is shown in debug mode only (or add a `verbose` keyword to omit the `@info` output)?
> * is this the correct place to put it? I somehow feel like this should go directly into `RadiationSpectra` applied to the output of `peakfinder` before returning it (?). This way, everybody using `RadiationSpectra.peakfinder` could profit from this.
> 
> Looking at LEGEND analysis:
> 
> * @theHenks would this interfere with finding the Tl208 FEP vs. finding the pulser peak if we sort them by ADC?

Hi, this for the feedback. This is work in progress and will not be merged like that into `co60`

PR #137, opened 2025-03-18
Rewrite `get_pe_res`: numerical FWHM calculation

### LINKED ISSUE #130: Determine resolution of P.E. peaks in SiPM gaussian mixture
Currently, the resolution of the P.E. peaks is calculated as
https://github.com/legend-exp/LegendSpecFits.jl/blob/6eece2dc01d1ac3612604ff4a3f9a50976650406/src/sipmfit.jl#L109-L111

This should be updated.

benedikt-nagler:
Fix #130.

The previous implementation of the `get_pe_res` function was incorrect. Since there is no closed formula for the FWHM of a gaussian mixture, it is now calculated numerically.

For each given subset of the gaussian mixture model, the maximum of the resulting PDF is calculated by dividing the interval [0.8, 1.2] into small steps. The step size depends on the largest variance of the individual gaussians, which ensures a high enough accuracy for the estimated maximum.

After that, all values where the half maximum is reached are calculated with the `find_zeros` function from `Roots.jl`. The FWHM then corresponds to the difference between the largest and the smallest position.

### DIFF
--- a/src/sipmfit.jl
+++ b/src/sipmfit.jl
@@ -109,9 +109,9 @@ function fit_sipm_spectrum(pe_cal::Vector{<:Real}, min_pe::Real=0.5, max_pe::Rea
     # get pe resolution (FWHM)
     get_pe_res = pe -> let sel = in.(μ, (-Δpe_peak_assignment..Δpe_peak_assignment) .+ pe)
         model_parameters = [
-            Vector(view(μ_ml, sel)),
-            Vector(view(σ_ml, sel)),
-            Vector(view(w_ml, sel))
+            view(μ_ml, sel),
+            view(σ_ml, sel),
+            view(w_ml, sel)
         ]
         
         # Find maximum of PDF in the expected range, step size depends on the maximum variance of the single gaussians
@@ -196,8 +196,8 @@ function _gmm_binned_loglike_func(
     return f_loglike
 end
 
-function _gmm_pdf(x::Real, μ::Vector{<:Real}, σ::Vector{<:Real}, w::Vector{<:Real})
+function _gmm_pdf(x::Real, μ::AbstractVector{<:Real}, σ::AbstractVector{<:Real}, w::AbstractVector{<:Real})
     return sum(@. w * exp(-0.5 * ((x - μ) / σ)^2) / (σ * sqrt(2π)))
 end
 
-_gmm_pdf(x::Real, V::Vector{<:Vector{<:Real}}) = _gmm_pdf(x, V...)
+_gmm_pdf(x::Real, V::Vector{<:AbstractVector{<:Real}}) = _gmm_pdf(x, V...)
--- a/src/sipmfit.jl
+++ b/src/sipmfit.jl
@@ -109,16 +109,17 @@ function fit_sipm_spectrum(pe_cal::Vector{<:Real}, min_pe::Real=0.5, max_pe::Rea
     # get pe resolution (FWHM)
     get_pe_res = pe -> let sel = in.(μ, (-Δpe_peak_assignment..Δpe_peak_assignment) .+ pe)
         model_parameters = [
-            view(μ_ml, sel),
-            view(σ_ml, sel),
-            view(w_ml, sel)
-        ]    
+            Vector(view(μ_ml, sel)),
+            Vector(view(σ_ml, sel)),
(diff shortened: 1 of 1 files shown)


fhagemann:
Unit tests?

src/sipmfit.jl
```diff
@@ -170,4 +193,10 @@ function _gmm_binned_loglike_func(
...
     return f_loglike
-end
\ No newline at end of file
+end
+
+function _gmm_pdf(x::Real, μ::Vector{<:Real}, σ::Vector{<:Real}, w::Vector{<:Real})
+    return sum(w .* exp.(-0.5 .* ((x .- μ) ./ σ).^2) ./ (σ .* sqrt(2π)))
+end
+
+_gmm_pdf(x::Real, V::Vector{<:Vector{<:Real}}) = _gmm_pdf(x, V[1], V[2], V[3])
```
fhagemann, line 202:
```suggestion
_gmm_pdf(x::Real, V::Vector{<:Vector{<:Real}}) = _gmm_pdf(x, V...)
```

src/sipmfit.jl
```diff
@@ -170,4 +193,10 @@ function _gmm_binned_loglike_func(
         xlogy.(bin_counts, binned_density)
     )
     return f_loglike
-end
\ No newline at end of file
+end
+
+function _gmm_pdf(x::Real, μ::Vector{<:Real}, σ::Vector{<:Real}, w::Vector{<:Real})
+    return sum(w .* exp.(-0.5 .* ((x .- μ) ./ σ).^2) ./ (σ .* sqrt(2π)))
```
fhagemann, line 199:
```suggestion
    return @. sum(w * exp(-0.5 * ((x - μ) / σ)^2) / (σ * sqrt(2π)))
```

src/sipmfit.jl
```diff
@@ -106,9 +105,33 @@ function fit_sipm_spectrum(pe_cal::Vector{<:Real}, min_pe::Real=0.5, max_pe::Rea
...
-        sqrt(dot(view(σ, sel).^2, view(w, sel).^2))
-    end 
+        model_parameters = [
+            view(μ_ml, sel),
+            view(σ_ml, sel),
+            view(w_ml, sel)
+        ]    
+        
+        # Find maximum of PDF in the expected range, step size depends on the maximum variance of the single gaussians
+        range_steps = maximum([μ[i].err for i in size(μ)]) / 100
```
fhagemann, line 118:
```suggestion
        range_steps = maximum(getfield.(μ, :err)) / 100
```
Check if this works. If not, use eachindex instead of size

fhagemann (changes requested):
Manchmal sieht man den Wald vor lauter Bäumen nicht 🥠🌳🔥

theHenks:
I am happy with the PR. Let's finish this soon and merge! 
Thanks a lot @benedikt-nagler 😍

src/sipmfit.jl
```diff
@@ -109,16 +109,17 @@ function fit_sipm_spectrum(pe_cal::Vector{<:Real}, min_pe::Real=0.5, max_pe::Rea
     # get pe resolution (FWHM)
     get_pe_res = pe -> let sel = in.(μ, (-Δpe_peak_assignment..Δpe_peak_assignment) .+ pe)
         model_parameters = [
-            view(μ_ml, sel),
-            view(σ_ml, sel),
-            view(w_ml, sel)
-        ]    
+            Vector(view(μ_ml, sel)),
+            Vector(view(σ_ml, sel)),
+            Vector(view(w_ml, sel))
```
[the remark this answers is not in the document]
benedikt-nagler: `view` returns the type `SubArray`. For the sake of consistency, I convert them to the `Vector` datatype.
This is relevant when calling the `_gmm_pdf` function, which I have only defined to take in `Vector` inputs.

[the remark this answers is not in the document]
fhagemann: You could also just loosen the `Vector` constrain in `_gmm_pdf` to `AbstractVector`.
`Vector` results in unnecessary allocation here

fhagemann (approved):
If this works without throwing an error, I would also be happy to merge.

theHenks:
> If this works without throwing an error, I would also be happy to merge.

I take this as a yes.

PR #140, opened 2025-03-20
Patch `CTC` energy

theHenks:
This PR increases the starting value to allow the `CTC` to converge easier. The earlier implementation could lead to non-converging trapping corrections such as: 
image

After the update, the same channel looks like:
image


fhagemann (approved):
This to me looks like a straight-forward fix

Issue #143, opened 2025-03-24
Non-normalized `LQ` cut classifier


theHenks:
The `LQ` cut classifier after the cut generation on the `DEP` is not normailzed to a µ = 0. and a σ = 1. Therefore, no detector comparison can be made. This has to be fixed by adjusting the output calibration func after the `CTC`.

PR #144, opened 2025-03-24
added lq_normalization

DaGeibl:
This pull reuest adds lq normalization. It adds a new lq_property function that normalizes the lq classifier to the DEP fit. The LQ cut then can be set at 3 of this normalized lq.

Something like this to fix issue https://github.com/legend-exp/LegendSpecFits.jl/issues/143?

The processor also needs to be updated to give the lq_cut function the right lq_classifier property function. I sadly can't quickly debug this, because something with my dataflow is broken, I try to debug my dataflow to verify this.


fhagemann:
Could we add tests here, please?

theHenks:
@DaGeibl Please also fix this in the `dataflow` so that this is consistent. We can then merge both PRs all together

DaGeibl:
@theHenks I now also added the pull request for the dataflow (https://github.com/legend-exp/legend-julia-dataflow/pull/92) and also verified that its running for period 1.  

@fhagemann what kind of test would you like?
I first thought about adding something like: 
    
    ` #Test the normalization
    # calculated normalization
    lq_norm = (lq_classifier_combined .- report_fit_result.μ ) ./ report_fit_result.σ
    @test isapprox(mean(lq_norm), 0.0, atol=0.05)
    @test isapprox(std(lq_norm), 1.0, atol=0.05) `
   
 But it is already kind of covered by the already existing tests

    `    # Test the parameters
    @test isapprox(report.fit_result.μ, expected_mean, atol=0.05)
    @test isapprox(report.fit_result.σ, expected_sigma, atol=0.05)
    @test isapprox(cut_3σ, expected_cut, atol=0.1)`

Because if they are true, the normalization tests also will be true.

DaGeibl:
I made a new commit. Is this how you wanted it?

theHenks:
What is the status here @DaGeibl ?

src/lqcut.jl
```diff
@@ -214,9 +214,13 @@ function lq_cut(
     # final cutoff value defined by "cut_sigma"
     cut_3σ = fit_result.μ + cut_sigma * fit_result.σ
 
+    # normalize lq classifier
+    lq_norm_func = " ( $(lq_class_expression)  - $(mvalue(fit_result.μ)) ) / ( $(mvalue(fit_result.σ)) )"
+
     result = (
         cut = cut_3σ,
         cut_fit_result = fit_result,
+        lq_norm_func = lq_norm_func,
```
fhagemann, line 223: I'm not sure if this matters, but isn't the convention to call the `ljl_propfunc` string just `func`?

test/test_lq.jl
```diff
@@ -84,4 +86,9 @@ end
     @test isapprox(report.fit_result.μ, expected_mean, atol=0.05)
     @test isapprox(report.fit_result.σ, expected_sigma, atol=0.05)
     @test isapprox(cut_3σ, expected_cut, atol=0.1)
+
+    # Test the normalization
+    lq_norm = (lq_classifier_peak1 .- report.fit_result.μ ) ./ report.fit_result.σ
+    @test isapprox(mean(lq_norm), 0.0, atol=0.05)
+    @test isapprox(std(lq_norm), 1.0, atol=0.05)
```
fhagemann, line 93: This is not necessarily testing the normalization you did in the function `lq_cut` but just showing that you could normalize the `lq_classifier` with the values in `result`.
Can you apply the `lq_norm_func` that is returned by `lq` to show that this actually does the correct thing?

fhagemann:
The failing tests should go away by themselves, after having merged #150.

src/lqcut.jl
DaGeibl: Ahh, didn't know/realize this. I updated this.

test/test_lq.jl
DaGeibl: ok, that was the simple solution. It took me a while to figure out how to make the property function work. 
But i updated it and now I am testing it properly

src/lqcut.jl
theHenks: Did you also change this in the dataflow?

test/Project.toml
```diff
@@ -16,4 +18,4 @@ Interpolations = "0.13, 0.14, 0.15"
 LegendDataTypes = "0.1"
 Measurements = "2"
 StatsBase = "0.33, 0.34"
-Unitful = "1"
\ No newline at end of file
+Unitful = "1"
```
fhagemann, line 21: I would recommend using the same `[compat]` entries as in the main `Project.toml`:
```suggestion
TypedTables = "1.2"
Unitful = "1.6"
```
and add `LegendDataManagement` in all versions for which `ljl_propfunc` was defined and working: `0.3, 0.4, 0.5`.

src/lqcut.jl
```diff
@@ -214,9 +214,13 @@ function lq_cut(
     # final cutoff value defined by "cut_sigma"
     cut_3σ = fit_result.μ + cut_sigma * fit_result.σ
 
+    # normalize lq classifier
+    lq_norm_func = " ( $(lq_class_expression)  - $(mvalue(fit_result.μ)) ) / ( $(mvalue(fit_result.σ)) )"
```
theHenks, line 218: I don't like the expression here without any brackets. The expression should be
``` julia
lq_norm_func = " ( ($(lq_class_expression)) - $(mvalue(fit_result.μ)) ) / ( $(mvalue(fit_result.σ)) )"
```
to make sure that there is never a math error here.
This was already an issue before, so I would like this to be changed and consistent.

test/Project.toml
theHenks, line 1: s. comments by Felix.

theHenks (changes requested):
We should be very careful with the expression handling. Let's rather have maybe some unnecessary brackets in the expressions rather then not having them.

test/Project.toml
fhagemann: image
Tests on downgrade are failing because it's pulling `LegendDataManagement@0.2.3` where it was not yet defined.

DaGeibl:
should i change the Project.toml then manually? What i did was: I activated the .toml and added the Packages through the Package manager. 
So since i probably need to add the compats manually, how do I in general find out wich are needed or should put in?

fhagemann:
> should i change the Project.toml then manually? What i did was: I activated the .toml and added the Packages through the Package manager. So since i probably need to add the compats manually, how do I in general find out wich are needed or should put in?

Yes, you would need to adjust the `[compat]` section manually.

How to choose which versions:
If the compats are already defined in the main `Project.toml` => take the same ones.
If not: check which versions allow for the functionality you are using and add them manually.
It might also be that the tests require newer versions that what the main `Project.toml` requires, then you should use the higher version as `[compat]`.

src/lqcut.jl
fhagemann:
```suggestion
    lq_norm_func = " ( $(lq_class_expression)  - $(mvalue(fit_result.μ)) ) / ( $(mvalue(fit_result.σ)) )"
```
```suggestion
    lq_norm_func = " ( ($(lq_class_expression))  - $(mvalue(fit_result.μ)) ) / ( $(mvalue(fit_result.σ)) )"
```

DaGeibl: Yes, I think i have the brackets in the dataflow side, so in the definition of lq_class_expression. But yeah it makes more sense to have them here and not in that dataflow

fhagemann: Better also having them here, just to be safe.

src/lqcut.jl
DaGeibl: No, on the dataflow I will be working next.

fhagemann (approved):

fhagemann (approved):
This should in principle be good to go, but I would wait until https://github.com/legend-exp/legend-julia-dataflow/pull/92 is also finished so these can go hand in hand.

DaGeibl:
Thanks Felix for addind the compat!

With my last pr i just moved a set of brackets from the dataflow processor to the `lq_ctc_correction` function. This bracket is not necessarly needed, but in my opinion adds clarity

DaGeibl:
As discussed with Florian, i reworked the lq_cut, to only normalize the lq classifier. The cut itself is performed with the normalized lq in a later stage in the la dataflow processor

src/lqcut.jl
```diff
@@ -109,8 +109,8 @@ function lq_ctc_correction(
     pol_fit_func = report_µ.f_fit
 
     # property function for drift time correction
-    lq_class_func = "$lq_e_corr_expression - " * join(["$(par[i]) * ($dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
-    lq_class_func_generic = "lq / e  - (slope * qdrift / e + y_inter)"
+    lq_class_func = "( $lq_e_corr_expression ) - " * join(["$(par[i]) * ($dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ")
```
fhagemann, line 112: Not sure if this is an overkill, but it doesn't harm to add a set of brackets around the whole expression here:
```suggestion
    lq_class_func = "( ( $lq_e_corr_expression ) - " * join(["$(par[i]) * ($dt_eff_expression)^$(i-1)" for i in eachindex(par)], " - ") * " )"
```

src/lqcut.jl
```diff
@@ -135,28 +135,29 @@ end
...
 # Keywords
-    * `cut_sigma`: Number of standard deviations used to define the final cutoff value
     * `dep_sideband_sigma`: Number of standard deviations used to define the sideband edges
     * `cut_truncation_sigma`: Number of standard deviations used for the precut of sideband subtracted histogram
+    * `uncertainty`: Boolean flag to include uncertainty in the fit (default: true)
+    * `lq_class_expression`: Expression for the used LQ classifier
 
 # Returns
-    * `result`: NamedTuple of the cutoff value
+    * `result`: NamedTuple of the normalization function
```
fhagemann, line 155: ... and also the fit result

src/lqcut.jl
DaGeibl: Mhm, they should be overkill, because i added in the `lq_norm` function brackets around the `lq_class_expression`:

`lq_norm_func = " ( ($(lq_class_expression))  - $(mvalue(fit_result.μ)) ) / ( $(mvalue(fit_result.σ)) )"`

So it think it would be overkill to have both brackets

theHenks (approved):

PR #147, opened 2025-03-29
add special handling for case when dof = 0

LisaSchlueter:
- Fixed p-value error in case when  `number fit parameter =  number of data points`
- example: linear energy calibration function for 2 Co-60 peaks
- previously error prevented  uncertainties to be calculated
- Now: `p-value = NaN`, but fit parameter uncertainties still provided

### DIFF
--- a/test/test_fit_chisq.jl
+++ b/test/test_fit_chisq.jl
@@ -30,5 +30,6 @@ using Test
     x = [1,2]
     y = f_lin.(x,par_true...) .+ 0.5.*randn(2)
     @info "chisq fit with 2 fit parameter on 2 data points (test of gof)"
-    result, report       = chi2fit(1, x, y; pull_t = [(mean = par_true[1], std= 0.1),(mean = par_true[2],std= 0.1)], uncertainty=true) 
+    @test_logs (:warn,) result, report       = chi2fit(1, x, y; pull_t = [(mean = par_true[1], std= 0.1),(mean = par_true[2],std= 0.1)], uncertainty=true) 
+    @test isnan(report.gof.pvalue) # check that the p-value is NaN
 end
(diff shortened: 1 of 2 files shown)


theHenks:
@LisaSchlueter Thanks for this PR. Could you quickly check if this change affects the recipes since the p-value is usually plotted there.

LisaSchlueter:
> @LisaSchlueter Thanks for this PR. Could you quickly check if this change affects the recipes since the p-value is usually plotted there.

Good point. Right now, the plot recipe shows " p = NaN" in the legend, which is not nice... 
https://github.com/legend-exp/LegendMakie.jl/pull/44 adds the option to remove the gof label for cases like that.

test/test_fit_chisq.jl
```diff
@@ -20,4 +20,9 @@ using Test
     y = f_lin.(x,par_true...) .+ 0.5.*randn(10)
     @info "chisq fit with uncertainties on x and y"
     result, report       = chi2fit(1, x, y; pull_t = [(mean = par_true[1], std= 0.1),(mean = par_true[2],std= 0.1)], uncertainty=true) 
+
+    x = [1,2]
+    y = f_lin.(x,par_true...) .+ 0.5.*randn(2)
+    @info "chisq fit with 2 fit parameter on 2 data points (test of gof)"
+    result, report       = chi2fit(1, x, y; pull_t = [(mean = par_true[1], std= 0.1),(mean = par_true[2],std= 0.1)], uncertainty=true)
```
fhagemann, line 27: Could you add a `@test` statement here.
Something like (just coded from my mind, check if it works)
```julia
@test_logs (:warn,) result, report = chi2fit(..., uncertainty = true) # to test that it throws your warning
@test isnan(report.gof.pvalue) # check that the p-value is NaN
```

theHenks (approved):
Perfect! Thanks @LisaSchlueter

theHenks:
@LisaSchlueter Are the comments from Felix all resolved? I am happy to merge this, but I would do this in parallel with the `LegendMakie.jl` PR so that everything is consistent.

LisaSchlueter:
> @LisaSchlueter Are the comments from Felix all resolved? I am happy to merge this, but I would do this in parallel with the `LegendMakie.jl` PR so that everything is consistent.

Yes, I implemented Felix suggestion. `LegendMakie.jl` Pr is merged too.

PR #148, opened 2025-03-29
Add minimal fit function for gamma peaks

LisaSchlueter:
- add new fit function `gamma_minimal` (only gauss + flat background) incl. pseudo-priors
- adjust fwhm calculation for case w/o low-energy tail
- improve tests


fhagemann (approved):
This looks good to me

theHenks (approved):
Thanks @LisaSchlueter

PR #149, opened 2025-04-01
Simple cal

LisaSchlueter:
- add a more generic simple_calibration function that can be accessed with keyword calib_type == :gamma which is useful for example for Co-60 data or other teststand data
- this is a more broken-down version of https://github.com/legend-exp/LegendSpecFits.jl/pull/108 (which will become obsolete with this PR)

- simple calibration report is plotable with modification in LegendMakie: https://github.com/legend-exp/LegendMakie.jl/pull/31/files

### DIFF
--- a/src/filter_optimization.jl
+++ b/src/filter_optimization.jl
@@ -26,6 +26,7 @@ function fit_enc_sigmas(enc_grid::Matrix{T}, enc_grid_rt::StepRangeLen{<:Quantit
     Threads.@threads for r in eachindex(enc_grid_rt)
         # get rt
         rt = enc_grid_rt[r]
+        
         # get enc for this rt
         enc_rt = flatview(enc_grid)[r, :]
         # sanity check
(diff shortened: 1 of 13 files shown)


src/filter_optimization.jl
```diff
@@ -26,6 +26,7 @@ function fit_enc_sigmas(enc_grid::Matrix{T}, enc_grid_rt::StepRangeLen{<:Quantit
     Threads.@threads for r in eachindex(enc_grid_rt)
         # get rt
         rt = enc_grid_rt[r]
+
```
theHenks, line 29: Could you remove the blank?

src/simple_calibration.jl
```diff
@@ -1,3 +1,4 @@
+const e_unit=u"keV" # unit for calibrated energies. All energy values will be convert to this unit
```
theHenks, line 1: Could you move this to the `LegendSpecFits.jl` file?

theHenks (changes requested):
Thanks a @LisaSchlueter

theHenks (approved):
Perfect! Thanks

PR #150, opened 2025-04-04
add exception when rt are out of range and all enc values are NaN

LisaSchlueter:
I ran into a few situations in which all `enc_rt` values for a particular `rt` are `NaN`. This can happen for very large `rt`.


theHenks:
@LisaSchlueter Could you check the failing tests?

theHenks:
> @LisaSchlueter Could you check the failing tests? 

Taklng it back. I think this is a bug in `GaussianMixtures`

theHenks (approved):

LisaSchlueter:
> > @LisaSchlueter Could you check the failing tests?
> 
> Taklng it back. I think this is a bug in `GaussianMixtures`

We had the same issue in `LegendMakie`. Do you want to wait until this (https://github.com/davidavdav/GaussianMixtures.jl/issues/121) is resolved or merge anyway?

theHenks:
> > > @LisaSchlueter Could you check the failing tests?
> > 
> > 
> > Taklng it back. I think this is a bug in `GaussianMixtures`
> 
> We had the same issue in `LegendMakie`. Do you want to wait until this ([davidavdav/GaussianMixtures.jl#121](https://github.com/davidavdav/GaussianMixtures.jl/issues/121)) is resolved or merge anyway?

I don't have strong opinions here. I let @fhagemann decide.

fhagemann:
The "clean" way to do it for now is to still run the job on `pre` (`julia-1.12`) but to allow for the tests to fail.

fhagemann:
I added a simple test which failed on the previous version and should now pass.
If it does, this should be free to merge.

fhagemann:
Downgrade tests fail because of something that was fixed in https://github.com/JuliaStats/StatsBase.jl/pull/792 and released in `StatsBase@0.33.17`.

fhagemann (approved):

PR #152, opened 2025-04-16
Exclude `Plots` extension from codecov

fhagemann:
A lot of untracked lines in codecov result from the `RecipesBase` extension to create plots with `Plots`.
To get a more realistic picture of the code coverage, I would exclude this extension from the codecov.


fhagemann:
For reference: the current codecov is 34%, whereas the codecov in just `src` is around 54%.
image

PR #153, opened 2025-04-16
Add tests for SiPM calibration routine

fhagemann:
I cannot judge how stable this is, so we'll see if there are some tests failing by chance.
This should cover a big chunk of the SiPM calibration code though.


fhagemann:
This PR is a mess...

PR #154, opened 2025-05-15
Reactivate tests on `pre`

fhagemann:
Now that a new version of `GaussianMixtures` was finally released, we can try reactivating the tests on julia-1.12 (`pre`)

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -27,8 +27,13 @@ jobs:
         run: julia -e 'using Pkg; Pkg.Registry.add("General"); Pkg.Registry.add(Pkg.RegistrySpec(url = "https://github.com/legend-exp/LegendJuliaRegistry"))'
         shell: bash
       - uses: julia-actions/julia-downgrade-compat@v1
+        if: matrix.version == 'pre'
+        with:
+          skip: Pkg,TOML,GaussianMixtures,Measurements,TypedTables,Unitful
+      - uses: julia-actions/julia-downgrade-compat@v1
+        if: matrix.version != 'pre'
         with:
           skip: Pkg,TOML,TypedTables
       - uses: julia-actions/julia-buildpkg@v1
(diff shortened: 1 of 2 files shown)


fhagemann:
Looks like `Unitful` does not compile on `julia-1.12` if the version is below 1.12.3 due to some definition of Angström.
I excluded downgrading `Unitful` for now on `pre` as well to prevent the tests from failing here.

theHenks:
> Looks like `Unitful` does not compile on `julia-1.12` if the version is below 1.12.3 due to some definition of Angström. [PainterQubits/Unitful.jl@559bd09](https://github.com/PainterQubits/Unitful.jl/commit/559bd09f5ad330ad055e5847bee3017ab2eda451) I excluded downgrading `Unitful` for now on `pre` as well to prevent the tests from failing here.

I don't understand this. Should this not just be a new unit? Why are our tests failing then?

fhagemann:
> > Looks like `Unitful` does not compile on `julia-1.12` if the version is below 1.12.3 due to some definition of Angström. [PainterQubits/Unitful.jl@559bd09](https://github.com/PainterQubits/Unitful.jl/commit/559bd09f5ad330ad055e5847bee3017ab2eda451) I excluded downgrading `Unitful` for now on `pre` as well to prevent the tests from failing here.
> 
> I don't understand this. Should this not just be a new unit? Why are our tests failing then?

Angström was defined before, but somehow but as `const` and `global`
Julia-1.12 doesn't like `const` and `global` at the same time.
So, the tests are failing because, when downgrading `Unitful` to the lowest version we support (1.6), `Unitful` as a whole package does not compile, and therefore also not `LegendSpecFits` because it depends on it.

I exclude DOWNGRADING `Unitful` in this PR FOR THE TESTS, so instead of using the lowest ALLOWED version on julia-1.12 (1.6.0) FOR THE TESTS, I let it choose the newest version it can FOR THE TESTS.
The package itself will not be affected by this.

PR #178, opened 2026-01-20
Fix chisq dof zero

jonasschlegel:
ErrorException("Error in LQ normalization calculation: DomainError(0.0, \"Chisq: the condition ν > zero(ν) is not satisfied.\")")


oschulz (approved):

jonasschlegel:
I will do this better, but open the PR for review anyway.

PR #193, opened 2026-08-06
Add fitting API components

oschulz:


tdixon97:
One feature that might be useful is passing a range to `binned_likelihood` ? what do you think?

oschulz:
> One feature that might be useful is passing a range to binned_likelihood ? what do you think?

You mean a range of bins? It might be cleaner to trim the histogram, then, maybe?

oschulz:
Btw, with this PR, BAT.jl >= v4.3.1 and PropertyFunctions.jl >= v0.3.2, we can now do this:

```julia
using StatsBase, Distributions, MeasureBase, DensityInterface, IntervalSets, BAT
using PropertyFunctions: @pf, PPath, input_property_paths as inputs
using LegendSpecFits
using LinearAlgebra: normalize
using Plots

peak1 = @pf batmeasure(Normal($mu1, $sigma))
peak2 = @pf batmeasure(Normal($mu2, $sigma))
intensity = @pf $a1 * peak1(_) + $a2 * peak2(_)

@show inputs(intensity)

prior = distprod(
    mu1 = 0.3..0.7, mu2 = 3..7, sigma = 0.5..2,
    a1 = 100..1000, a2 = 500..5000
)

v_truth = (mu1 = 0.5, mu2 = 5.0, sigma = 1.0, a1 = 400, a2 = 1000)
m = intensity(v_truth)

A = float(massof(m))
N_data = rand(Poisson(A))
events = rand(m^N_data)

h = fit(Histogram, events, -4:0.2:8)
plot(h)

L = binned_likelihood(intensity, h)
logdensityof(L, v_truth)
r = bat_bgml(L, prior)
v_bgml, _ = r

plot(normalize(h, mode = :density), label = "observed")
plot!(-4:0.05:8, densityof(intensity(v_truth)), lw = 4, label = "truth")
plot!(-4:0.05:8, densityof(intensity(v_bgml)), lw = 2, label = "fit")
```

The parameters of `peak1` and `peak2` automatically become part of the parameters of `intensity` here (new special semantics of `f(_)` calls inside of `@pf`).

oschulz:
CC @wiesinger-ch

tdixon97:
> > One feature that might be useful is passing a range to binned_likelihood ? what do you think?
> 
> You mean a range of bins? It might be cleaner to trim the histogram, then, maybe?

Yeah this is what I mean, I think trimming the histogram works but (if theres not another way I dont know) requires a few steps and either modifying the `Histogram` or producing a new one.

I guess it would be something like:

```julia
data = rand(1000)*100
h = fit(Histogram, data, 0:1:100)

# Keep only bins from 20 to 80
mask = (h.edges[1][1:end-1] .>= 20) .& (h.edges[1][2:end] .<= 80)

new_edges = h.edges[1][findfirst(mask):findlast(mask)+1]
new_weights = h.weights[mask]

h_trim = Histogram((new_edges,), new_weights)
```
I think it would certainly be nice to vary the range of bins used in the likelihood without having to go through these steps. I am especially thinking about very new users who maybe do not understand the internal structure of a `Histogram` (and do not really need to).

I would find it very nice to be able to do something like:

```julia
L = binned_likelihood(expectation, h,bin_range = (20,50))
```
but I realise this would require some changes (maybe the trimming shown above).

tdixon97:
Another unrelated question. When I look at your examples the models are defined by PDFs, and then inside `binned_likelihood` you handle the integration. 
But we have cases where our models are already binned (for example template) fits, then it might make more sense to supply directly the "forward model", i.e the function computing the expectation value in each bin.

I wonder if also in the case that your model is based on some PDF / measure / distribution (sorry I'm unsure the right term), it would be more natural to split the integration and the likelihood definition?

theHenks:
@oschulz I am wondering if this should not rather become part of the upstream `RadiationSpectra.jl` ? There, we could formalize the API and also add recipes for Makie etc. 
And in a second step, refactor all `LegendSpecFits.jl` code to this new API.

oschulz:
> I am wondering if this should not rather become part of the upstream RadiationSpectra.jl

Yes, we'll definitely want to upstream this to RadiationSpectra and BinnedModels. But those two packages are both due for an overhaul first, and some of that depends on other things that are currently in flight. So the plan is to incubate this functionality in LegendSpecFits first and then upstream a bit later. This way people can start using it right away.

theHenks:
> > I am wondering if this should not rather become part of the upstream RadiationSpectra.jl
> 
> Yes, we'll definitely want to upstream this to RadiationSpectra and BinnedModels. But those two packages are both due for an overhaul first, and some of that depends on other things that are currently in flight. So the plan is to incubate this functionality in LegendSpecFits first and then upstream a bit later. This way people can start using it right away.

I see. However, I could at least see some of the functionality to be upstreamed already, such as the shape measures etc. 

Independent, we should also soonish plan to rework the rest of the fitting code in LegendSpecFits to move it to this new API. I would like to avoid different codes performing fitting within the same package, which makes it unmaintainable over time.

PR #194, opened 2026-08-18
sipm_simple_calibration: valley-based noise cut, noise threshold in result and report

jonasschlegel:
## What didn't work before
The noise/1PE cut of `sipm_simple_calibration` was a fixed offset above the noise peak
(`cuts_1pe.max + n_fwhm_noise_cut * fwhm_noise`). On channels where the noise tail blends
into the 1PE peak, or where the valley is wide and empty, that cut landed inside the noise
or the 1PE peak, the peak search then locked onto the wrong maximum, and the threshold that
was actually used was not part of the result — so it could neither be stored nor plotted.

## What works now
- **Valley detection:** the noise/1PE threshold is found at the centroid of the first valley
  after the noise peak on a smoothed 40-bin histogram (`_find_noise_threshold`);
  `n_fwhm_noise_cut` caps the search window in units of the noise-peak half-width,
  `0.0` skips detection (uses `initial_min_amp`), negative values keep the legacy formula.
  Split 1PE peaks are merged before the peak search.
- **Peak search:** the left window starts at the minimum after the noise peak; trigger logic
  for the VoV input reworked, optional `single_trigger_only` second-stage QC keeps only
  waveforms with exactly one trigger above the detected threshold.
- **Result / report:** `result` gains `noise_threshold` and `noise_threshold_cal`; `report`
  gains `noise_threshold*`, `valley_found` and the unfiltered all-trigger spectra
  `h_uncal_full` / `h_calsimple_full` for plotting.
- **Perf:** the cut pool is built with `flatview` instead of pairwise `vcat`.

## Downstream
LegendMakie recipe (composite plot with the noise cut marker) and Juleana (persist the
threshold per detector in the SiPM pars) — PRs linked below.

### DIFF
--- a/src/sipm_simple_calibration.jl
+++ b/src/sipm_simple_calibration.jl
@@ -26,6 +26,15 @@ Returns
 function sipm_simple_calibration end
 export sipm_simple_calibration
 
+function sipm_simple_calibration(pe_uncal_vov::VectorOfVectors{<:Real}; initial_min_amp::Real=0.0, initial_max_amp::Real=50.0, relative_cut_noise_cut::Real=0.5, n_fwhm_noise_cut::Real=5.0, kwargs...)
+    pe_flat = filter(isfinite, reduce(vcat, pe_uncal_vov))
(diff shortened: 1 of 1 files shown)


jonasschlegel:

theHenks (approved):

PR #197, opened 2026-08-18
Fits must not crash or hang on states the optimizer only probes: poisson_binll, estimate_fwhm scan cap

jonasschlegel:
- **`poisson_binll(λ, k)`** (new, used in the likelihoods and in `gof`): returns `-Inf` for
  negative or non-finite `λ` instead of throwing. During a fit the optimizer routinely
  probes parameter sets with a negative or NaN model density (negative background amplitude,
  width → 0); their log-likelihood is `-Inf` and the optimizer moves on, whereas
  `Poisson(λ)` raised a `DomainError` that aborted the whole fit.
- **`estimate_fwhm`**: the FWHM scan step is `max(1e-3, range/1e5)` — an MC draw of the fit
  parameters with a huge σ made the fixed 1e-3 scan run for hours (this was the hanging
  ecal step in the production); physical widths keep the 1e-3 step exactly.

### DIFF
--- a/src/gof.jl
+++ b/src/gof.jl
@@ -74,7 +74,7 @@ function p_value_poissonll(fit_func::Base.Callable, h::Histogram{<:Real,1}, v_ml
     counts, bin_widths, bin_centers = _prepare_data(h) # prepare data
     model_func = Base.Fix2(fit_func, v_ml) # fix the fit parameters to ML best-estimate
 
-    bin_ll(x, bw, k) = logpdf(Poisson(bw * model_func(x)), k) # loglikelihood per bin, evaluated for best-fit parameters (v_ml)
+    bin_ll(x, bw, k) = poisson_binll(bw * model_func(x), k) # loglikelihood per bin, evaluated for best-fit parameters (v_ml)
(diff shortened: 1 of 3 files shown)


jonasschlegel:

jonasschlegel:
Claude comment on failing test: 

The only failing job is Julia pre (the Julia 1.13 pre-release build). The failure is not related to this PR: GenericLinearAlgebra (a dependency of BAT via DoubleFloats) overwrites a LinearAlgebra stdlib method, and Julia 1.13 no longer permits method overwriting during precompilation — so the package fails to load before any test runs. This hits every branch (and main) the same way and has to be fixed upstream in GenericLinearAlgebra. All stable-Julia jobs (Julia 1 on Linux/macOS/Windows, Julia min) are green.

jonasschlegel:
How should we fix this correctly? @oschulz. The fit function is returning negative PDF values, so the fit is crashing directly from `Poisson(λ)` if the likelihood returns `-inf`. This is a hot fix I think (a check for negative parameters like background etc..). There already was a comment in line 19 in `likelihoods.jl` already.

jonasschlegel:
poisson_binll(λ, k) this fix is outdated and the estimate_fwhm() fix will be opened in an new PR.

PR #198, opened 2026-08-18
Better error and debug messages

jonasschlegel:
No change to successful fits — only what happens when a channel cannot be fitted:
- **`fit_fwhm_ft` / `fit_sf_wl`:** record the per-gridpoint skip reason (empty peak window,
  invalid peak-stats guess, < 100 events, SF outside (1, 100) %) and surface the counts in
  the error when no (resp. fewer than 2) grid points survive, with a hint (check the FEP
  peak / DEP-SEP statistics of this channel; det-specific override or usability change).
  `fit_sf_wl` now requires ≥ 2 valid points — a single point is no optimization and
  downstream report consumers assume a grid.
- **`cut_single_peak`:** clear error on an empty energy window; a peak flush against the
  window edge falls back to the last edge instead of `nothing + Int`.
- **`fit_calibration` / `fit_fwhm` / `peakstats`:** speaking `ArgumentError`s when no or too
  few peaks passed the fit QC (`Vector{Any}`/`Vector{Union{}}` inputs, `sort(peaks)[1:2]`
  BoundsError, sideband windows overlapping the peak with ≤ 14 bins).


jonasschlegel:

PR #199, opened 2026-08-26
Support ArraysOfArrays v1

oschulz:
Adds ArraysOfArrays v1 to the compat bounds, keeping support for the previous versions. All ArraysOfArrays API used here (`flatview` on arrays of similar arrays and on plain matrices, `ArrayOfSimilarArrays{<:Real}` / `VectorOfVectors{<:Real}` in signatures) is unchanged in v1.

CI can exercise ArraysOfArrays v1 only once EncodedArrays 0.5.0, LegendDataTypes 0.1.16 and BAT 4.3.2 are registered; until then the resolver keeps picking v0.6.

Includes a patch version bump.

Created by generative AI.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -57,7 +57,7 @@ LegendSpecFitsRecipesBaseExt = ["RecipesBase", "Plots"]
 
 [compat]
 ArgCheck = "1, 2"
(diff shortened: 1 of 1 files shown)


oschulz:
Closes #200.

PR #202, opened 2026-09-07
Peak-maximum search robustness (FWHM err) and correct resolution-curve start values (FWHM fit)

### LINKED ISSUE #156: Update priors of fit_fwhm
The fit priors of the `fit_fwhm` defined in `get_fit_fwhm_pseudo_prior` should be updated, as the fit sometimes does not converge for detectors with bad energy resolution:

Image

jonasschlegel:
Hi, this PR has two FWHM fixes with one commit each. 

The first fixes the FWHM(E) fit `fit_fwhm` with setting the correct **squared** prior for the intercept from a weighted least squared fit instead of taking the (not squared) intercept calculated with the first 2 fwhm values (the fit is FWHM^2(E) and not FWHM(E). This resolves Issue: [https://github.com/legend-exp/LegendSpecFits.jl/issues/156](https://github.com/legend-exp/LegendSpecFits.jl/issues/156)

The second change is making the error calculation of the FWHM more robust. 

**Summary and details by Claude:** 

`fit_fwhm:` the start value and prior mode of the constant term were a FWHM intercept in keV, but the parameter is fwhm² in keV² — 3–15× too low for FWHM ≳ 3 keV, and since the prior saturates at ~4× its mode, the fit got stuck at the guess (worst cases: χ²/ndf ~1000, FWHM at Qbb off by up to 1 keV). Now both start values and the prior mode come from a weighted least squares of fwhm² = enc + fano·E; model, likelihood and prior shape unchanged. On 457 detector sets no fit gets stuck anymore, FWHM at Qbb matches a prior-free reference to 0.005 keV.

`estimate_fwhm:` the 0.001 keV step scan for the peak maximum stalls on MC draws from unconverged fits (σ up to ~1e6 keV) and returns the bracket edge if the maximum lies outside. Replaced by bisection on the derivative: bounded steps, NaN as before if there's no maximum in the bracket. Same FWHM to 3e-7 keV over 828 production peaks.


jonasschlegel:
The wrong prior `fit_fwhm` only affected bad detectors with a high intercept, because of the quadratic prior. Her are some example plots before and after the fix: 

Good Detector: 
l200-p16-r000-cal-det02-fwhm_e_cusp_ctc_before_after

Bad Detector:

l200-p16-r000-cal-det04-fwhm_e_cusp_ctc_before_after

jonasschlegel:
1. benchmark fwhm error with derivative 
2. check fano factor prior --> tighter than 0.3 
3. check chi^2 fit --> squared fit 
4. check error on Qbb with scaled reduced xi^2

theHenks:
- Check alternative chi2 optimizers, e.g. NativeMinuit.jl and the threaded options in there
- Rebase onto my improvements branch
- Check if FWHM extraction with more peaks becomes more robust

PR #207, opened 2026-09-16
Refactor peak sf

DaGeibl:
Hi,
this PR refactors the `get_peaks_survival_fractions` to make it more general, as I want to be able to use not only for `lq` or `aoe`, but input a generic cut vector (that could be e.g. a QC flag) and generate the survival fraction for that.

I also refactored the old code that used a confusing "inverted" keyword. Instead the PSD survival flags are now created by a new `get_su`rvival_flag fucntion, that can have both a high or low cut, or just one of them.

I also added that the sf is now saved in %. Please comment if that's good, or should be removed.
`result_after.sf * 100u"percent"`

Also @theHenks: Do we want to keep the deprecation of `surrival`  or can we safely remove that?

Codex summary:
Simplifies LegendSpecFits survival-fraction calculations around a Boolean selection, with optional low and high cuts. Updates the affected JuLeana A/E, LQ, and PSD processor calls, and preserves the expected result and report fields. Adds focused tests for peak and continuum selections, including report histograms. All 25 focused tests and the dispatch-ambiguity check pass.


theHenks:
Can you give some code examples?

DaGeibl:
The changes in Yeah i forgot to open this PR :sweat_smile: 

The code example for the QC is found here:
https://github.com/legend-exp/Juleana.jl/pull/182

src/aoe_cut.jl
```diff
@@ -85,233 +85,146 @@ end
 export get_low_aoe_cut
 
 
+function get_survival_flag(cut_parameter::AbstractVector{<:Unitful.RealOrRealQuantity};
```
theHenks: This seems not needed and could go to one of the dispatch functions.

src/aoe_cut.jl
```diff
@@ -85,233 +85,146 @@ end
...
+Fit a peak before and after an arbitrary event selection. The histogram method accepts
+count histograms for all, surviving, and rejected events with identical, unitless
+energy binning; the latter two must partition the first. It returns fit results and
+reports without a nominal `peak` label.
+
+The event-selection method constructs these histograms and adds the supplied `peak`
+to both outputs. The interval-based method constructs the selection and forwards
+to the event-selection method. Bounds are exclusive.
+"""
+function get_peak_survival_fraction(e::AbstractVector{<:T}, peak::T, window::AbstractVector{T}, survival_flag::AbstractVector{Bool};
```
theHenks: Maybe restrict the window to be a Tuple, or a vector of length 2?

theHenks: Also I am missing the "normal" sf mode:
``` julia
get_peak_survival_fraction(e, psd_flag)
```
where I have the energy values of my peak and a `psd_falg` bool vector.

theHenks (changes requested):

theHenks (changes requested):
Could you please use:
- A consistent set of variable names and declaration
- A consistent typing across all functions and dispatches (e.g. `Unitful.Energy{<:Real}` or `MaybeWithEnergyUnits` and so on. 
- A consistent typing for windows. I think they should be a `Tuple` for now, should become `Intervals` later. But this is for a refactor
- Histograms should have some types. 
- There is a huge mix between `Abstract` types and normal types.
- The docstring should represent the types too.

theHenks (approved):

